Issue #16, opened 2019-02-16
Add option to set custom timezone


JianWang2016:
Hi @Aiden:

This is an amazing package with very clear documentation.  It would be great to have option to set custom timezone.  

Thanks much.

JianWang2016:
This is just perfect!

PR #21, opened 2020-04-03
add --timeZoneOffset argument for cwa cwa.gz files; expose csv arguments to python script

xiangnandang:
Hi Aiden @aidendoherty,

Thanks for creating such an excellent package for the community. This package has been greatly helpful for my projects.

I have created an additional argument --timeZoneOffset (only for cwa cwa.gz files for now), to address this issue https://github.com/activityMonitoring/biobankAccelerometerAnalysis/issues/16, where device configuration and device deployment are conducted in different timezone.

for instance, if the device is configured in US eastern time zone, and deployed in US pacific time zone, "python3 accProcess.py data/sample.cwa.gz --timeZoneOffset -180" would do the job. "-180" refers to when the device deployment local time is 3 hours (180 minutes) behind the configure local time.

The second commit I submitted was simply to surface the csv arguments from java script to python script.

Please let me know if you need more information from me. I'm looking forward to incorporating these features to benefit more users, as well looking forward to developing more features/enhancements for this excellent package.

Best,
Xiangnan (dang_xiangnan@lilly.com)

### DIFF
--- a/accProcess.py
+++ b/accProcess.py
@@ -35,7 +35,8 @@ def main():
                             metavar='e.g. -180', default=0,
                             type=int, help="""timezone minutes offset induced 
                             by timezone difference from configure timezone and
-                            deployment timezone""")
+                            deployment timezone
+                            (default : %(default)s""")
     parser.add_argument('--startTime',
                             metavar='e.g. 1991-01-01T23:59', default=None,
                             type=str2date, help="""removes data before this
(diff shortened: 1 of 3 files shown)


aidendoherty:
Many thanks @xiangnandang !! This is very helpful. Please do feel free to suggest further updates and improvements in the future, this is much appreciated.

PR #23, opened 2020-04-19
fix minor issue in accUtils.loadTimeSeriesCSV for --timeSeriesDataCol…

xiangnandang:
…umn True

Hi Aiden @aidendoherty,

there's a minor issue in the package: if --timeSeriesDateColumn True, accPlot.py will raise error due to the timestamp column (in the timeseries.csv file) was not converted to datetime type and set as index in accUtils.loadTimeSeriesCSV function.

I'm submitting this pull request to fix this issue by adding these two lines in accUtils.loadTimeSeriesCSV:
else:
        tsData = tsData.set_index(pd.to_datetime(tsData[TIME_SERIES_COL]))

Best,
Xiangnan


aidendoherty:
Good spot, many thanks @xiangnandang  !

PR #24, opened 2020-04-20
Gt3v2 support round 1

angerhang:
Added incomplete support for GT3V2

1. Detection of v1 and v2 for `.gt3` format.
2. V2 packet header parsing.
3. Checksum validation for each packet.
4. Several minor code beautifications. 

To test this, run `AxivityAx3Epoches` on any valid gt3 v2 file.

### DIFF
--- a/java/AxivityAx3Epochs.java
+++ b/java/AxivityAx3Epochs.java
@@ -15,6 +15,7 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.RandomAccessFile;
 import java.math.RoundingMode;
+import java.lang.Math;
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
 import java.nio.channels.Channels;
@@ -37,7 +38,8 @@ import java.time.ZoneOffset;
 import java.text.DecimalFormat;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
-
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 /**
  * Calculates epoch summaries from an AX3 .CWA file. Class/application can be
@@ -56,6 +58,15 @@ public class AxivityAx3Epochs {
 
 	private static Boolean verbose = true;
 	private static EpochWriter epochWriter;
+	private static Logger logger;
+
+	static {
+		String path = AxivityAx3Epochs.class.getClassLoader()
+				.getResource("logging.properties")
+				.getFile();
+		System.setProperty("java.util.logging.config.file", path);
+		logger = Logger.getLogger(AxivityAx3Epochs.class.getName());
+	}
 
 	/*
 	 * Parse command line args, then call method to identify and write epochs.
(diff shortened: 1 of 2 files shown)


angerhang:
Additional features include:

1. A class level logger for debugging, which can be potentially used for all other pipelines as well. Currently, it is only used to process GT3 v2 files.
2. Parameter reading and value decoding from floating-point numbers.

java/AxivityAx3Epochs.java
```diff
@@ -156,6 +165,7 @@ public static void main(String[] args) {
 				} else if (funcName.equals("timeZoneOffset")) {
 					timeZoneOffset = Integer.parseInt(funcParam);
 				} else if (funcName.equals("verbose")) {
+					logger.setLevel(Level.ALL);
```
aidendoherty, line 168: Excuse my ignorance of using a class logger, but is there a possibility the log file can create conflicts when we do multi-CPU processing on the cluster?

Is there any easy way to toggle the logger on/off?

angerhang: Yes, this can be easily turned off by setting the level to `none`. More references here https://docs.oracle.com/javase/8/docs/api/java/util/logging/class-use/Level.html. In terms of potential conflict with the cluster processing, I am not sure as I am not familiar with how the multi-CPU processing is done. 

Have a logger for the V2 support is really helpful because we are processing lots of smaller packets. Being able to toggle different levels of logging is really useful for debugging. Maybe the best way is to remove the logger level for the `verbose` mode and set the level to None for the default mode? 

We can update this after we make sure that there is no conflict with the cluster processing?

aidendoherty: Yes, that sounds good. So can we go with Level.OFF as the default - and then switch logging on if we go with Verbose=True ?

angerhang: Just set the default mode to `off` and one just change the logging level by modifying the `logging.properties` file https://github.com/activityMonitoring/biobankAccelerometerAnalysis/pull/24/commits/bcfdc2d093ffd5b27a0a3ed16abee133bdec4f3a

angerhang:
Sorry for the hassle, but I just have just added the working version of Gt3v2 support, being able to convert GT3 v2 file into npy. But it seems a bit slow. I am running this over tonight to see if we can get a sensible `npy` output in the morning :D mostly the changes are from this commit https://github.com/activityMonitoring/biobankAccelerometerAnalysis/pull/24/commits/1ad03c5ade1c3c2bf17c0429397ffae8496cb103

java/AxivityAx3Epochs.java
aidendoherty: Many thanks. Before merging this, let's just wait unto you are happy with the changes made last night. Great work!

angerhang: Think it should be ready to go. The timestamp had some issues with the unit but has been corrected. The npy file seems reasonable. :D One can just test via: `java AxivityAx3Epochs test.gt3x npyOutput:true npyFile:test.npy`

aidendoherty: Many thanks @angerhang  - great work!

PR #35, opened 2020-05-02
handle low sample rates in sanDiegoFFT

chanshing:
This is to fix an issue when passing `sampleRate` less than 30, in which case we cannot compute FFT0-FFT14 in `sanDiegoFFT`. To handle this, just resample the signal to 30Hz right before computing.


chanshing:
I figured I will close this and open a new PR since the change is now in `Features.java` after the major code refactoring. Please see https://github.com/activityMonitoring/biobankAccelerometerAnalysis/pull/38

PR #38, opened 2020-05-04
handle sampleRate<30 in sanDiegoFFT

chanshing:
This is to fix an issue when passing `sampleRate` less than 30, in which case we cannot compute FFT0-FFT14 in sanDiegoFFT. To handle this, just resample the signal to 30Hz right before computing.


chanshing:
Closing this in favor of https://github.com/activityMonitoring/biobankAccelerometerAnalysis/pull/39

PR #42, opened 2020-05-21
Arm angel

angerhang:
Add two arm-angel features, `avgArmAngel`, `avgArmAngelAbsDiff` to assist sleep vs awake classification based on 

* Estimating sleep parameters using an accelerometer without sleep diary (van Hees et al. 2018) 
* A novel, open access method to assess sleep duration using a wrist-worn accelerometer (van Hees et al. 2015)

The computed features have been tested on the provided sample file that yields the following results:

Screen Shot 2020-05-21 at 1 27 47 PM

`avgArmAngel` has been verified to be between -90 and 90 and `avgArmAngelAbsDiff` is non-negative.


aidendoherty:
Great work @angerhang !

PR #45, opened 2020-05-26
fix acc scale setting for gt3x v1

angerhang:
Fix setting the default value for `accelerationScale`  when `info.txt` doesn't contain one.


angerhang:
Merged into `arm_angel` branch for the sleep project usage, thus closing.

PR #47, opened 2020-05-30
allow display of activity classification during imputed period of time

xiangnandang:
…imputed period of time

### DIFF
--- a/accPlot.py
+++ b/accPlot.py
@@ -129,13 +129,14 @@ def plotTimeSeries(
         # and then plot time series data for this day
         plt.subplot(nrows, 1, i+1)
         plt.plot(timeSeries, group['acc'], c='k')
-        plt.fill(timeSeries, np.multiply(group['imputed'], ymax),
-            labels_as_col['imputed'], alpha=1.0)
+        plt.fill_between(timeSeries, y1 = np.multiply(group['imputed'], ymax), 
+            y2 = np.multiply(group['imputed'], ymax * 0.5),
+            color = labels_as_col['imputed'], alpha=1.0)
 
         # change display properties of this subplot
         ax = plt.gca()
         if len(labels)>0:
-            ax.stackplot(timeSeries, [np.multiply(group[l], ymax) for l in labels],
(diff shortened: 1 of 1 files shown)


xiangnandang:
Hi,

This pull request is to address this requirement in the output plots: during imputed period, the only information readable from the plot is that this period of time is imputed, but not about the imputed activity classification.

The current pull request provides a naive solution to use roughly bottom 50% to display activity classification for both imputed and non-imputed period of time (just the entire time), and use the top 50% to highlight the imputed period. The outcome is as this graph (showing one day and the early morning period is imputed):

image

During imputed period, the relative bar graph height indicate the possibility of activity classification as originally designed, noting the probabilities not always add to 1, could this be a rounding issue during imputation for activity classification?

I hope this helps in general to display both imputation flag as well as imputation result in the output. The 50% bottom and 50% top could either be changed to a better default, or supplied as an argument (I don't think that's quite necessary).

Best,
Xiangnan

chanshing:
Thanks for your contribution. I am not sure about this. @aidendoherty @R-Walmsley @angerhang what do you think?

xiangnandang:
@chanshing , thanks for your comment. I absolutely see that this option should not be the default, so maybe add as an argument to allow advanced users to see more details into imputed periods?

Best,
Xiangann

angerhang:
I think the option is nice to have but maybe too wasteful in terms of space usage. Maybe we should change the ratio to 80%-20% or 70%-30%.

xiangnandang:
@angerhang , I agree about the ratio. I'll later make it a true/false argument for the call, with default as false, and set the the ratio to 25%.

xiangnandang:
added --imputedLabels as an argument to control this behavior, and default to False, plotting as original:
image

if --imputedLabels True, then plot with top 25% area as indicator for imputation, and bottom 75% area to plot imputed activity classification 
image

Best,
Xiangnan

aidendoherty:
This code is very useful. Many thanks @xiangnandang !
Also many thanks @chanshing and @angerhang for your comments and feedback - it is very much appreciated.

PR #48, opened 2020-06-02
allow daylight savings crossover check from different timezones

xiangnandang:
This pull request is to allow daylight savings crossover check from different timezones.

Currently, in function of check_daylight_savings_crossover, it is hardcoded to check for timezone "Europe/London". This commit is to set up a new argument of "timeZone" to allow more flexibility.

The default of the "timeZone" has been set to "Europe/London" to allow compatibility. This argument has been exposed to accProcess.py 's parser, function of getActivitySummary, and function of check_daylight_savings_crossover.

(the commit history was a bit messed up as I started a new branch in the middle of another pull request, but the files changed are clean.)

Best,
Xiangnan

### DIFF
--- a/accPlot.py
+++ b/accPlot.py
@@ -44,10 +44,6 @@ def main():
                             metavar='True/False', default=True, type=str2bool,
                             help="""Highly recommended method to show imputed 
                             missing data (default : %(default)s)""")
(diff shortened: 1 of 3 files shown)


aidendoherty:
Many thanks for this improvement @xiangnandang - very much appreciated!

PR #51, opened 2020-06-06
timeZone for check_daylight_savings_crossover

xiangnandang:
a fresh pull request to address timeZone argument for check_daylight_savings_crossover function in summariseEpoch.py

@aidendoherty, please take a look, I have added step-by-step comment/explanation.

To be honest, dealing with timezone and dst is quite challenging, and I hope this could be addressing the challenge. Here (also in the second comment) is what I think the pain point is:

The key challenge in this work is that all timestamps in e.index are defined without timezone information, meaning that when a timestamp is defined as 5pm, it's supposed to refer to 5pm local time, however the timestamp does not have timezone association, in effect in "UTC"

However, the output of pytz._utc_transition_times are the dst crossover times defined in UTC, for example, while most dst crossover happens at 2am, the "America/New_York" (UTC-5) 's _utc_transition_times are 7am during spring crossover (7am UTC = 2am US/eastern without dst), and 6am during fall crossover (6am UTC = 2am US/eastern with dst). To address this, a series of transformation is required on the output of _utc_transition_times, the variable "transition" here.

### DIFF
--- a/accelerometer/summariseEpoch.py
+++ b/accelerometer/summariseEpoch.py
@@ -202,6 +202,7 @@ def check_daylight_savings_crossover(e, startTime, endTime, summary,
     """
 
     daylightSavingsCrossover = 0
+    utc_tz = pytz.timezone("UTC")
     localTime = pytz.timezone(timeZone)
     # Convert because pytz can error if not using python datetime type
     startTimeZone = localTime.localize(startTime.to_pydatetime())
@@ -218,11 +219,14 @@ def check_daylight_savings_crossover(e, startTime, endTime, summary,
         for t in localTime._utc_transition_times:
             if t>startTime:
                 transition = t
+                transition = transition - pd.DateOffset(seconds = 1)
+                transition = utc_tz.localize(transition)
+                transition = localTime.normalize(transition)
+                transition = transition.replace(tzinfo = None)
+                transition = transition + pd.DateOffset(seconds = 1)
                 break
         # If Autumn crossover time, adjust transition time plus remove 1hr chunk
         if offset == -1:
-            # pytz stores dst crossover at 1am, but clocks change at 2am local
-            transition = transition + pd.DateOffset(hours=1)
             # Remove last hr before DST cut, which will be subsequently overwritten
             e = e[(e.index < transition - pd.DateOffset(hours=1)) |
                     (e.index >= transition)]


accelerometer/summariseEpoch.py
```diff
@@ -202,6 +202,7 @@ def check_daylight_savings_crossover(e, startTime, endTime, summary,
     """
 
     daylightSavingsCrossover = 0
+    utc_tz = pytz.timezone("UTC")
```
xiangnandang, line 205: helper timezone object to allow "localize" into "UTC"

accelerometer/summariseEpoch.py
```diff
@@ -218,11 +219,14 @@ def check_daylight_savings_crossover(e, startTime, endTime, summary,
         for t in localTime._utc_transition_times:
             if t>startTime:
                 transition = t
+                transition = transition - pd.DateOffset(seconds = 1)
```
xiangnandang, line 222: The key challenge in this work is that all timestamps in e.index are defined without timezone information, meaning that when a timestamp is defined as 5pm, it's supposed to refer to 5pm local time, however the timestamp does not have timezone association, in effect in "UTC"

However, the output of pytz._utc_transition_times are the dst crossover times defined in UTC, for example, while most dst crossover happens at 2am, the "America/New_York" (UTC-5) 's _utc_transition_times are 7am during spring crossover, and 6am during fall crossover. To address this, a series of transformation is required on the output of _utc_transition_times, the variable "transition" here.

The first step is a "hack" that dial the "transition" back for 1 second. This is because, during spring crossover, the same UTC transition time, 7am (in UTC) for US eastern, corresponds to both 2am and 3am local time, as they are essentially the same time. By dialing back 1 second, 6:59:59am UTC will deterministically become 1:59:59am US/eastern.

accelerometer/summariseEpoch.py
```diff
@@ -218,11 +219,14 @@ def check_daylight_savings_crossover(e, startTime, endTime, summary,
...
+                transition = utc_tz.localize(transition)
```
xiangnandang, line 223: In addition, even "transition" is meant to describe timestamp in UTC, it's not associated with any timezone. Here, it's just using the earlier defined "utc_tz" to give "transition" a "UTC" timezone association, so that additional timezone handling can be performed next.

accelerometer/summariseEpoch.py
```diff
@@ -218,11 +219,14 @@ def check_daylight_savings_crossover(e, startTime, endTime, summary,
...
+                transition = localTime.normalize(transition)
```
xiangnandang, line 224: "normalize" the transition from UTC to local timezone

the outcome of this line, for US/eastern:
1. spring crossover: 6:59:59am -> 1:59:59am US/eastern
1. fall crossover: 5:59:59am -> 1:59:59am US/eastern

accelerometer/summariseEpoch.py
```diff
@@ -218,11 +219,14 @@ def check_daylight_savings_crossover(e, startTime, endTime, summary,
...
+                transition = transition.replace(tzinfo = None)
```
xiangnandang, line 225: in the last line, the output is almost what we wanted, meaning the value of 1:59:59am is correct (1 second apart from 2am crossover), but we want to remove the timezone association, so that it can be used with comparison / calculation with e.index timestamps

calling "replace" with tzinfo=None would do exactly that.

the outcome of this line, for US/eastern:
1. spring crossover: 1:59:59am US/eastern -> 1:59:59am
1. fall crossover: 1:59:59am US/eastern -> 1:59:59am

accelerometer/summariseEpoch.py
```diff
@@ -218,11 +219,14 @@ def check_daylight_savings_crossover(e, startTime, endTime, summary,
...
+                transition = transition + pd.DateOffset(seconds = 1)
```
xiangnandang, line 226: now dialing forward the 1 second we dialed back (the "hack")

the outcome of this line, for US/eastern:
1. spring crossover: 1:59:59am -> 2am
1. fall crossover: 1:59:59am -> 2am

accelerometer/summariseEpoch.py
```diff
@@ -218,11 +219,14 @@ def check_daylight_savings_crossover(e, startTime, endTime, summary,
...
                 break
         # If Autumn crossover time, adjust transition time plus remove 1hr chunk
         if offset == -1:
-            # pytz stores dst crossover at 1am, but clocks change at 2am local
-            transition = transition + pd.DateOffset(hours=1)
```
xiangnandang, line 225: since the previous steps have already resulted in correct output (2am) for both spring and fall crossover, this step is no longer necessary

chanshing:
Thanks @xiangnandang 
I have tested your changes and it works fine.

aidendoherty:
Many thanks @xiangnandang for the very thorough explanation on this. 
I also tested these changes and they work great. Many thanks again.

PR #54, opened 2020-06-12
fix small bug: missing field in summary.json

chanshing:
Smallest PR ever.

This is to fix a missing field in the summary.json output file `quality-daylightSavingsCrossover` which was only being set when there was a crossover.


aidendoherty:
Great spot @chanshing!

PR #55, opened 2020-06-15
Native DST crossover correction for CWA files

chanshing:
So this is to fix the issue with the timestamps in epoch.csv files not being corrected for daylight savings time (DST) crossover. A half-baked solution would be to just do the correction in summariseEpoch.py, which was being done already but not re-saved to reflect the correction (only timeSeries.csv reflected the change). However, a better long term solution is to do the DST correction natively in the Java processing pipeline, so that whatever files come out are already corrected so that we don't need to do it one by one with the possibility of missing some (for example, it is not only the epoch.csv that had this issue but also the nonWearBouts.csv, stationaryPoints.csv, raw csv and npy files -- basically everything that comes out of the Java processing part).

The main changes with this PR are that now the DST correction is done at the source in `readCwaBuffer()`, and output files (epoch.csv, nonWearBouts.csv, etc.) are DST-corrected and zoned to user-specified timezone (default: Europe/London). The timeSeries.csv file is also changed to have a time column, which is needed because the times are now zoned and offsets may change in between.

The PR seems to contain a lot of changes in parts due to how intertwined parts of the code are, but it just mostly consists of timezone localization/conversion/formatting. I did my best to split the work into commits of manageable sizes and added comments below for clarity. Please review my PR and feel free to suggest modifications! @aidendoherty @R-Walmsley @angerhang


java/AxivityReader.java
```diff
@@ -16,6 +18,10 @@
  */
 public class AxivityReader extends DeviceReader {
 
+    // public static ZonedDateTime SESSION_START = null;
+    // public static ZoneId zoneId = ZoneId.of("Europe/London");
+    public static LocalDateTime SESSION_START = null;
```
chanshing: `SESSION_START` is persisted throughout the cwa file reads. This is the _only_ reliable local time that we have as any subsequent timestamp may have suffered a DST crossover.

java/AxivityReader.java
```diff
@@ -220,8 +226,16 @@ private static void readCwaBuffer(ByteBuffer buf, LocalDateTime SESSION_START,
                 }
 
                 // determine time for indexed sample within block
-                LocalDateTime blockTime = getCwaTimestamp((int) blockTimestamp, 
+                LocalDateTime blockTime = getCwaTimestamp((int) blockTimestamp,
                                                     fractional, timeZoneOffset);
+
+                // if SESSION_START not set yet, this is the first block
```
chanshing: As comment says, if `SESSION_START` is not set yet, then this is the first block in the file read so it contains the start time.

accelerometer/accUtils.py
```diff
@@ -94,9 +94,10 @@ def generateTimeSeries(epochPD, tsFile, timeSeriesDateColumn=False,
 
     startTime = pd.to_datetime(epochPD.index.values[0])
     endTime = pd.to_datetime(epochPD.index.values[-1])
-    # resample index so we can add imputed data
-    e = epochPD.reindex(pd.date_range(start=startTime, end=endTime,
-        freq=str(epochPeriod)+'s'))
```
chanshing, line 99: For some reason reindexing the _zoned_ times makes all other columns NaNs. Not reindexing has worked equally fine here.

accelerometer/summariseEpoch.py
```diff
@@ -13,13 +13,13 @@
 from datetime import timedelta
 
 def getActivitySummary(epochFile, nonWearFile, summary,
```
chanshing: Nothing to see here. This is just a quirk of my editor (VSCode) that automatically removes trailing spaces (a best practice) in the file.

accelerometer/summariseEpoch.py
```diff
@@ -104,9 +104,6 @@ def getActivitySummary(epochFile, nonWearFile, summary,
     # Get interrupt and data error summary vals
     e = get_interrupts(e, epochPeriod, summary)
```
chanshing, line 109: We no longer need to correct for DST crossovers here

accelerometer/summariseEpoch.py
```diff
@@ -160,12 +157,11 @@ def get_interrupts(e, epochPeriod, summary):
     """
 
     epochNs = epochPeriod * np.timedelta64(1, 's')
```
chanshing, line 166: Slightly simplify these

accelerometer/summariseEpoch.py
```diff
@@ -180,71 +176,6 @@ def get_interrupts(e, epochPeriod, summary):
...
-        e['newTime'] = np.where(e['newTime'].isnull(), e.index, e['newTime'])
-        e = e.set_index('newTime')
-        # Reset startTime and endTime variables
-        startTime = pd.to_datetime(e.index.values[0])
-        endTime = pd.to_datetime(e.index.values[-1])
-    # And record to output summary
-    summary['quality-daylightSavingsCrossover'] = daylightSavingsCrossover
-    return e
-
-
```
chanshing, line 247: We no longer need this

accelerometer/summariseEpoch.py
```diff
@@ -180,71 +176,6 @@ def get_interrupts(e, epochPeriod, summary):
     return e
 
 
-def check_daylight_savings_crossover(e, startTime, endTime, summary,
```
chanshing, line 183: We no longer need this

java/AxivityReader.java
```diff
@@ -18,19 +20,25 @@
  */
 public class AxivityReader extends DeviceReader {
 
-    // public static ZonedDateTime SESSION_START = null;
-    // public static ZoneId zoneId = ZoneId.of("Europe/London");
-    public static LocalDateTime SESSION_START = null;
```
chanshing: Now `SESSION_START` is a zoned time and its DST status `sessionStartDST` will be registered. `sessionStartDST` is important as all subsequent timestamps' DST status compared against to see if a crossover has occurred.

java/AxivityReader.java
```diff
@@ -312,14 +322,16 @@ private static void readCwaBuffer(ByteBuffer buf,
                     x = xRaw / 256.0;
                     y = yRaw / 256.0;
                     z = zRaw / 256.0;
-                    timeValue = getEpochMillis(blockTime);
```
chanshing: This is where the correction is called. `blockTime` which is a `LocalDateTime` object (no timezone) is zoned with DST correction (`zonedWithDSTCorrection()`), then converted to Unix time in milliseconds (milliseconds since 1 Jan 1970 UTC).

java/AxivityReader.java
```diff
@@ -312,14 +322,16 @@ private static void readCwaBuffer(ByteBuffer buf,
...
+                    t = zonedWithDSTCorrection(blockTime).toInstant().toEpochMilli();
 
-                    epochWriter.newValues(timeValue, x, y, z, temperature, errCounter);
+                    epochWriter.newValues(t, x, y, z, temperature, errCounter);
 
                 }
             } catch (Exception excep) {
                 excep.printStackTrace(System.err);
-                System.err.println("block err @ " + lastBlockTime[0].toString() + ": " + excep.toString());
```
chanshing, line 316: For printing purposes, zone and DST-correct the time. Note that internally the code still works with `LocalDateTime` objects because of some low-level processes.

java/AxivityReader.java
```diff
@@ -348,5 +360,25 @@ private static LocalDateTime cwaHeaderLoggingStartTime(ByteBuffer buf) {
...
+    private static void setTimeZone(String timeZone) {
+        zoneId = ZoneId.of(timeZone);
+        rules = zoneId.getRules();
+    }
+
+
+    private static void setSessionStart(LocalDateTime ldt) {
+        sessionStart = ldt.atZone(zoneId);
+        sessionStartDST = rules.isDaylightSavings(sessionStart.toInstant());
+    }
```
chanshing: These are called when we start reading a CWA file. They set the timezone and register the DST status of the session start time.

java/AxivityReader.java
```diff
@@ -348,5 +360,25 @@ private static LocalDateTime cwaHeaderLoggingStartTime(ByteBuffer buf) {
...
+
+
+    private static ZonedDateTime zonedWithDSTCorrection(LocalDateTime ldt) {
+        ZonedDateTime zdt = ldt.atZone(zoneId);
+        boolean isDST = rules.isDaylightSavings(zdt.toInstant());
+        if (isDST != sessionStartDST) {  // DST crossover happened
+            zdt = isDST ? ldt.plusHours(1).atZone(zoneId) : zdt.plusHours(-1);
+        }
+        return zdt;
+    }
```
chanshing, line 382: At last, the DST crossover correction code. It takes a local time (no timezone), zones it to the user-specified timezone (e.g. Europe/London), checks to see if DST has changed (wrt. session start), and if so correct accordingly.

It is perhaps worth noting why forward and backward shifts are done differently. When stepping _into_ DST `ZonedDateTime` is smart enough to automatically push forward local times that lie in the inexistent hour, for example `LocalDateTime.parse("2019-03-31`**`T01:05`**`").atZone(ZoneId.of("Europe/London")` automagically produces `2019-03-31T`**`02:05+01:00`**`[Europe/London]` because push forward happened at 01:00 and so 01:05 doesn't really exist. This is great functionality from `ZonedDateTime`, but the same will not happen with hours that may well exist, e.g. `LocalDateTime.parse("2019-03-31T`**`02:05`**`").atZone(ZoneId.of("Europe/London")`  also produces the same timestamp `2019-03-31T`**`02:05+01:00`**`[Europe/London]`. To solve this we just manually push forward the local hour _before_ zoning.

When stepping _out of_ DST, a different situation happens. For example `LocalDateTime.parse("2019-10-27`**`T01:05`**`").atZone(ZoneId.of("Europe/London")` is really an ambiguous time (is this 01:05 _before_ the push back at 02:00 or _after_?), but the output of such line is `2019-10-27T`**`01:05+01:00`**`[Europe/London]`, i.e. `ZonedDateTime` prefers to keep the DST status before the pushback. So using `LocalDateTime` there's no way to obtain something like `2019-10-27T`**`01:05+00:00`**`[Europe/London]` i.e. 01:05 _after_ push back. To solve this we manually push back  _after_ the zoning, so to obtain the above time we first do `LocalDateTime.parse("2019-10-27`**`T02:05`**`").atZone(ZoneId.of("Europe/London")` which first produces `2019-10-27T`**`02:05+00:00`**`[Europe/London]`, then `.plusHours(-1)` to produce `2019-10-27T`**`01:05+00:00`**`[Europe/London]`.

java/EpochWriter.java
```diff
@@ -243,11 +257,12 @@ public boolean newValues(
 	 *  [the above does not apply if getSanDiegoFeatures is enabled]
 	 */
 	private void writeEpochSummary(
-			LocalDateTime epochStartTime,
+			// ZonedDateTime epochStartTime,
+			Instant epochStartTime,
```
chanshing: Instead of taking a `LocalDateTime` object, take an `Instant` object (represents a specific point in time).

accelerometer/accUtils.py
```diff
@@ -94,9 +94,10 @@ def generateTimeSeries(epochPD, tsFile, timeSeriesDateColumn=False,
...
+    # # resample index so we can add imputed data
+    # e = epochPD.reindex(pd.date_range(start=startTime, end=endTime,
+    #     freq=str(epochPeriod)+'s'))
+    e = epochPD
```
chanshing, line 100: For some reason reindexing the zoned times makes all other columns NaNs. Not reindexing has worked equally fine here. I think reindexing was being done as a way to find gaps and do imputation, but maybe this is no longer needed as this is already done before.

accelerometer/summariseEpoch.py
```diff
@@ -13,13 +13,13 @@
...
-    epochPeriod=30, stationaryStd=13, minNonWearDuration=60, 
-    mgMVPA=100, mgVPA=425, 
+    epochPeriod=30, stationaryStd=13, minNonWearDuration=60,
+    mgMVPA=100, mgVPA=425,
     activityModel="activityModels/doherty-may20.tar",
     intensityDistribution=False, useRecommendedImputation=True,
-    psd=False, fourierFrequency=False, fourierWithAcc=False, m10l5=False, 
+    psd=False, fourierFrequency=False, fourierWithAcc=False, m10l5=False,
     verbose=False):
     """Calculate overall activity summary from <epochFile> data
```
chanshing: Nothing to see here. This is just a quirk of my editor (VSCode) that automatically removes trailing spaces (a best practice) in the file.

accelerometer/summariseEpoch.py
```diff
@@ -104,9 +104,6 @@ def getActivitySummary(epochFile, nonWearFile, summary,
     # Get interrupt and data error summary vals
     e = get_interrupts(e, epochPeriod, summary)
 
-    # Check if data occurs at a daylight savings crossover
-    e = check_daylight_savings_crossover(e, startTime, endTime, summary, timeZone)
-
```
chanshing, line 109: We no longer need to correct for DST crossovers here

accelerometer/summariseEpoch.py
chanshing, line 247: We no longer need all this

java/AxivityReader.java
```diff
@@ -18,19 +20,25 @@
...
 
-    // public static ZonedDateTime SESSION_START = null;
-    // public static ZoneId zoneId = ZoneId.of("Europe/London");
-    public static LocalDateTime SESSION_START = null;
+    private static ZonedDateTime sessionStart = null;
+    private static boolean sessionStartDST;
+
+    private static ZoneId zoneId;
+    private static ZoneRules rules;
+
```
chanshing: Now `SESSION_START` (renamed `sessionStart` for consistency) is a zoned time and its DST status `sessionStartDST` will be registered.` sessionStartDST` is important as all subsequent timestamps' DST status is compared against to see if a crossover has occurred.

accPlot.py
```diff
@@ -101,7 +101,10 @@ def plotTimeSeries(
     """
 
     # read time series file to pandas DataFrame
-    d = accUtils.loadTimeSeriesCSV(tsFile)
+    d = pd.read_csv(
+        tsFile, index_col='time',
+        parse_dates=['time'], date_parser=accUtils.date_parser
+    )
```
chanshing, line 107: No longer need specific loading function required to parse through the column names to determine the time column

accProcess.py
```diff
@@ -352,12 +352,6 @@ def deleteIntermediateFiles():
         fourierWithAcc=args.fourierWithAcc, m10l5=args.m10l5,
         verbose=args.verbose)
 
-    # Generate time series file (note: this will also resample to epochData so do this last)
-    accelerometer.accUtils.generateTimeSeries(epochData, args.tsFile,
-        epochPeriod=args.epochPeriod,
-        timeSeriesDateColumn=args.timeSeriesDateColumn,
-        activityClassification=args.activityClassification, labels=labels)
-
```
chanshing, line 366: This part has now been moved into summariseEpoch.py

accelerometer/accUtils.py
```diff
@@ -502,3 +503,26 @@ def createDirIfNotExists(folder):
...
+
+
+
+def date_strftime(t):
+    '''
+    Convert to time format of the form e.g.
+    2020-06-14 19:01:15.123000+0100 [Europe/London]
+    '''
+    tz = t.tz
+    return t.strftime(f'%Y-%m-%d %H:%M:%S.%f%z [{tz}]')
```
chanshing, line 528: These functions will be generally useful to parse through the time format of the output files from the Java pipeline

accelerometer/summariseEpoch.py
```diff
@@ -79,8 +80,10 @@ def getActivitySummary(epochFile, nonWearFile, summary,
         e = epochFile
     else:
         # Use python PANDAS framework to read in and store epochs
-        e = pd.read_csv(epochFile, parse_dates=['time'], index_col=['time'],
-            compression='gzip').sort_index()
+        e = pd.read_csv(
+            epochFile, index_col=['time'],
+            parse_dates=['time'], date_parser=accUtils.date_parser,
+        )
```
chanshing, line 86: We need special `date_parser` to parse through the time format that now contains timezones.

accelerometer/summariseEpoch.py
```diff
@@ -423,3 +430,29 @@ def writeMovementSummaries(e, labels, summary, useRecommendedImputation):
...
+
+    e_new = pd.DataFrame(index=e.index)
+    e_new['imputed'] = e.isna().any(1).astype('int')
+    e_new[cols_new] = e[cols]
+
+    # make output time format contain timezone
+    # e.g. 2020-06-14 19:01:15.123000+0100 [Europe/London]
+    e_new.index = e_new.index.to_series().apply(accUtils.date_strftime)
+
+    e_new.to_csv(tsFile, compression='gzip')
```
chanshing: New code that essentially replaces `accUtils.generateTimeSeries()` to write out `timeSeries.csv` with a new format: it now has a time column containing zone info.

java/AccelerometerParser.java
```diff
@@ -52,8 +52,10 @@ public static void main(String[] args) {
     	DF3.setRoundingMode(RoundingMode.HALF_UP); // To match their implementation
 
     	// epochWriter parameters
-    	int epochPeriod = 30;
-    	DateTimeFormatter timeFormat = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss.SSS");
+        int epochPeriod = 30;
+        // output time format, e.g. 2020-06-14 19:01:15.123000+0100 [Europe/London]
+        // this should be consistent with the date_parser used later in the Python code
+        final DateTimeFormatter timeFormat = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss.SSSSSxxxx '['VV']'");
```
chanshing, line 58: Time format to use in the csv output files from the Java pipeline. As comment says, `2020-06-14 19:01:15.123000+0100 [Europe/London]`
This needs to match the `date_parser` function used later in the Python pipeline.

java/AccelerometerParser.java
```diff
@@ -114,9 +116,6 @@ public static void main(String[] args) {
 					verbose = Boolean.parseBoolean(funcParam.toLowerCase());
 				} else if (funcName.equals("epochPeriod")) {
 					epochPeriod = Integer.parseInt(funcParam);
-				// Make sure time f
-				} else if (funcName.equals("timeFormat")) {
-					timeFormat = DateTimeFormatter.ofPattern(funcParam);
```
chanshing, line 118: Remove the ability to specify a time format as this will mess up with Panda's `date_parser`

chanshing:
@aidendoherty @R-Walmsley @angerhang

java/AxivityReader.java
```diff
@@ -306,22 +322,24 @@ private static void readCwaBuffer(ByteBuffer buf, LocalDateTime SESSION_START,
                     x = xRaw / 256.0;
                     y = yRaw / 256.0;
                     z = zRaw / 256.0;
-                    timeValue = getEpochMillis(blockTime.plusNanos(START_OFFSET_NANOS));
+                    t = zonedWithDSTCorrection(blockTime).toInstant().toEpochMilli();
```
chanshing, line 325: This is where the correction is called. blockTime which is a LocalDateTime object (no timezone) is first zoned with DST correction (see zonedWithDSTCorrection()), then converted to Unix time in milliseconds (milliseconds since 1 Jan 1970 UTC). NOW it is truly a Unix time!

aidendoherty:
Hi @chanshing 

Many thanks for the very helpful documentation. I completely agree with you that a better long term solution is to do the DST correction natively in the Java processing pipeline - great idea!

I have some high level comments and questions on this pull request, and will also try to add further comments to each of your commit notes above:

1) Can you move the Java timeZone updates to the EpochWriter.Java file?
This then means that this code is just needed one time, and can be used for .gt3x, .bin, and .csv input files too. This would certainly be helpful for the .gt3x and .bin files we process as these also store localDateTime.

2) Will this work for a scenario where the device is setup in one timezone, but deployed in another? i.e. should we keep the timeZoneOffset integer in our code?
@xiangnandang - perhaps you might like to comment on this?

3) The -timeSeries.csv.gz file intentionally did not write out the datetime column to save disk space. However, I imagine your reason for now including this column is to explicitly handle files which happen during a timeZone change?

4) Can you add/keep some code to retain the 'quality-daylightSavingsCrossover' flag in the -summary.json file ?

5) Is there a reason to move the -timeSeries.csv.gz writing code to within the summariseEpoch.py file?
To me this more logically fits as a utility function rather than something necessary to derive summary measures from the input epoch file.

java/AxivityReader.java
aidendoherty: This makes me think that it would be better to conduct timeZone corrections at the EpochWriter.java level.

java/AxivityReader.java
aidendoherty: Can you edit the text in this comment to move from the somewhat confusing (to me at least!) 'push forward'/'push back' terminology to 'spring forward'/'fall back' (more commonly used to describe daylight saving changes)?

I love this comment, and we need a way to integrate it into the wider documentation for this tool (I'll look into this) - this is really helpful.

accelerometer/summariseEpoch.py
aidendoherty: No problem, at some point I will apply a linter to the project

accelerometer/summariseEpoch.py
aidendoherty: The only part I'd like to keep is the 'quality-daylightSavingsCrossover' flag, if possible.

aidendoherty: Oh and my reason for this is that it would be helpful in subsequent health association analysis for sleep and circadian rhythm use cases i.e. we may like to remove such people from our analysis (assuming their sleep/circadian status is abnormally changed during a daylight saving crossover week)

accProcess.py
aidendoherty: Is there a reason to move the -timeSeries.csv.gz writing code to within the summariseEpoch.py file?

To me this more logically fits as a utility function rather than something necessary to derive summary measures from the input epoch file.

accelerometer/summariseEpoch.py
aidendoherty: Again, I have a preference to keep this as a utility function. However, I there may be a better reason for moving it to summariseEpoch.py?

java/AxivityReader.java
chanshing: Thanks Aiden, I improved the comments as per the suggestions. I hope it is clearer now.

java/AxivityReader.java
chanshing: Perhaps it would be good to perform the correction at the very source instead of delegating to subsequent processes? Maybe EpochWriter.java should just receive the values all corrected and be in charge of just writing/formatting.
Also, I think `lastBlockTime` would still need to be DST-corrected for printing anyway.

accelerometer/summariseEpoch.py
chanshing: Ahhh right, that should be maintained. I will submit a fix soon.

accelerometer/summariseEpoch.py
chanshing: Ok I can move this out of summariseEpoch. I will submit a fix soon :)

accProcess.py
chanshing: Thanks Aiden for taking the time to review the changes!

> 1. Can you move the Java timeZone updates to the EpochWriter.Java file?
>    This then means that this code is just needed one time, and can be used for .gt3x, .bin, and .csv input files too. This would certainly be helpful for the .gt3x and .bin files we process as these also store localDateTime.

I was thinking maybe it is preferable to perform the correction at the source instead of delegating this to subsequent processes and keep EpochWriter just in charge of writing/formatting. But I haven't considered .gt3x and .bin files and I'm not too familiar with them. Can you help verify that this is correct https://github.com/actigraph/GT3X-File-Format ? If the device's internal representation of time is in Unix time then it won't need DST correction.
Generally, I think because devices have their own format it makes semantic sense to have _their_ side of the parsing code handle what's necessary to output the correct values. What do you think?

> 2. Will this work for a scenario where the device is setup in one timezone, but deployed in another? i.e. should we keep the timeZoneOffset integer in our code?
>    @xiangnandang - perhaps you might like to comment on this?

Ah I see, in that case I will recover this functionality. But perhaps we should change its name to something like `timeShift` or `timeOffset`? Because timezone offset would be a bit misleading: If a device was configured in NY (UTC-5) and shipped to UK summertime (UTC+1) then the device needs +6 hours. `--timeZoneOffset 6` I think would be confusing.

> 3. The -timeSeries.csv.gz file intentionally did not write out the datetime column to save disk space. However, I imagine your reason for now including this column is to explicitly handle files which happen during a timeZone change?

Yes, because there can be DST crossovers and the offsets would change, e.g. if we're in Italy then the column might go `2020-03-29`**`01:59:00 +0100`**, `2020-03-29`**`01:59:30+0100`**, `2020-03-29`**`03:00:00+0200`**, ...

> 4. Can you add/keep some code to retain the 'quality-daylightSavingsCrossover' flag in the -summary.json file ?

Yes I will submit a fix soon. Thanks for spotting this.

> 5. Is there a reason to move the -timeSeries.csv.gz writing code to within the summariseEpoch.py file?
>    To me this more logically fits as a utility function rather than something necessary to derive summary measures from the input epoch file.

That makes sense, I will go back to the original behavior.

chanshing:
Thanks Aiden for taking the time to review the changes!

> 1. Can you move the Java timeZone updates to the EpochWriter.Java file?
>    This then means that this code is just needed one time, and can be used for .gt3x, .bin, and .csv input files too. This would certainly be helpful for the .gt3x and .bin files we process as these also store localDateTime.

I was thinking maybe it is preferable to perform the correction at the source instead of delegating this to subsequent processes and keep EpochWriter just in charge of writing/formatting. But I haven't considered .gt3x and .bin files and I'm not too familiar with them. Can you help verify that this is correct https://github.com/actigraph/GT3X-File-Format ? If the device's internal representation of time is in Unix time then it won't need DST correction.
Generally, I think because devices have their own format it makes semantic sense to have _their_ side of the parsing code handle what's necessary to output the correct values. What do you think?

> 2. Will this work for a scenario where the device is setup in one timezone, but deployed in another? i.e. should we keep the timeZoneOffset integer in our code?
>    @xiangnandang - perhaps you might like to comment on this?

Ah I see, in that case I will recover this functionality. But perhaps we should change its name to something like `timeShift` or `timeOffset`? Because timezone offset would be a bit misleading: If a device was configured in NY (UTC-5) and shipped to UK summertime (UTC+1) then the device needs +6 hours. `--timeZoneOffset 6` I think would be confusing.

> 3. The -timeSeries.csv.gz file intentionally did not write out the datetime column to save disk space. However, I imagine your reason for now including this column is to explicitly handle files which happen during a timeZone change?

Yes, because there can be DST crossovers and the offsets would change, e.g. if we're in Italy then the column might go `2020-03-29`**`01:59:00 +0100`**, `2020-03-29`**`01:59:30+0100`**, `2020-03-29`**`03:00:00+0200`**, ...

> 4. Can you add/keep some code to retain the 'quality-daylightSavingsCrossover' flag in the -summary.json file ?

Yes I will submit a fix soon. Thanks for spotting this.

> 5. Is there a reason to move the -timeSeries.csv.gz writing code to within the summariseEpoch.py file?
>    To me this more logically fits as a utility function rather than something necessary to derive summary measures from the input epoch file.

That makes sense, I will go back to the original behavior.

java/AxivityReader.java
```diff
@@ -330,18 +348,37 @@ private static LocalDateTime getCwaTimestamp(int cwaTimestamp, int fractional, i
...
 
-    private static LocalDateTime cwaHeaderLoggingStartTime(ByteBuffer buf, int timeZoneOffset) {
+    private static LocalDateTime cwaHeaderLoggingStartTime(ByteBuffer buf) {
         long delayedLoggingStartTime = getUnsignedInt(buf, 13);
-        return getCwaTimestamp((int) delayedLoggingStartTime, 0, timeZoneOffset);
+        return getCwaTimestamp((int) delayedLoggingStartTime, 0);
+    }
+
+
+    private static void setTimeZone(String timeZone) {
```
aidendoherty, line 363: @chanshing - can you move this to the abstract DeviceReader.java class?
i.e. we would like this to be available to all of AxivityReader.java, ActigraphReader.java, etc.

java/AxivityReader.java
```diff
@@ -330,18 +348,37 @@ private static LocalDateTime getCwaTimestamp(int cwaTimestamp, int fractional, i
...
+        zoneId = ZoneId.of(timeZone);
+        rules = zoneId.getRules();
     }
 
 
+    private static void setSessionStart(LocalDateTime ldt) {
```
aidendoherty, line 369: @chanshing - can you move this to the abstract DeviceReader.java class?
i.e. we would like this to be available to all of AxivityReader.java, ActigraphReader.java, etc.

java/AxivityReader.java
```diff
@@ -330,18 +348,37 @@ private static LocalDateTime getCwaTimestamp(int cwaTimestamp, int fractional, i
...
+        sessionStart = ldt.atZone(zoneId);
+        sessionStartDST = rules.isDaylightSavings(sessionStart.toInstant());
+    }
+
+
+    private static ZonedDateTime zonedWithDSTCorrection(LocalDateTime ldt) {
```
aidendoherty, line 375: @chanshing - can you move this to the abstract DeviceReader.java class?
i.e. we would like this to be available to all of AxivityReader.java, ActigraphReader.java, etc.

java/AxivityReader.java
aidendoherty: After further reflection, I agree with you @chanshing

java/AxivityReader.java
aidendoherty: @angerhang - we may have to do this for the ActigraphReader.java class too?

java/AxivityReader.java
```diff
@@ -16,16 +20,25 @@
  */
 public class AxivityReader extends DeviceReader {
 
+    private static ZonedDateTime sessionStart = null;
```
aidendoherty, line 23: @chanshing - can you move this block of 4 variable declarations to the abstract DeviceReader.java class?

i.e. I'm pretty sure we would like this to be available to all of AxivityReader.java, ActigraphReader.java, etc.

java/AxivityReader.java
```diff
@@ -7,7 +7,11 @@
 import java.nio.channels.FileChannel;
 import java.nio.channels.ReadableByteChannel;
 import java.time.Duration;
+import java.time.LocalDate;
 import java.time.LocalDateTime;
+import java.time.ZoneId;
```
aidendoherty, line 12: @chanshing - do you think we can move these new import statements to the abstract DeviceReader.java class?

i.e. we would like these to be available to all of AxivityReader.java, ActigraphReader.java, etc.

aidendoherty:
Hi @chanshing 

On further reflection - I agree that you are right to do the timezone assignment at source i.e. within AxivityReader.java rather than EpochWriter.java. I've made a few comments on moving some parts of the code into our abstract DeviceReader.java file. Once you have made these, I can then accept this pull request and merge it into the master directory.

Many thanks for your help and advice on this.

java/EpochWriter.java
```diff
@@ -124,9 +134,9 @@ public EpochWriter(
 	// Returns true to continue processing, or false if endTime has been reached
 	public boolean newValues(
```
aidendoherty, line 135: @chanshing - can you add in a comment here to very explicitly state what the variable "long time" expects?

chanshing:
Hi @aidendoherty I applied the changes that you suggested in the latest 3 commits. Basically:

1. Restored original behavior of when to write timeSeries.csv
2. Restored original behavior of flagging daylight savings crossovers in the summary.json
3. Restored time shifting functionality
4. Refactor timezone and DST correction code -- these bits are now in DeviceReader

For point 4, it would be worth checking if the other devices really need DST correction, because if their internal time is in Unix time or UTC then there's no need.

I hope this helps. Thank you.

accPlot.py
aidendoherty: Hi @chanshing - given we will no longer support it, can you delete the loadTimeSeriesCSV() method from accUtils.py ?

chanshing:
Hi @aidendoherty thanks for the suggestions, I cleaned up accUtils.py in the last commit. Thanks!

aidendoherty:
Looks great, many thanks @chanshing !

PR #56, opened 2020-06-18
handle potential nan in correlation features

chanshing:
Fix potential nan values due to precision error and/or degenerate random variables.

@aidendoherty I hope this solves the issue of correlations tending to be nan.

### DIFF
--- a/java/AccStats.java
+++ b/java/AccStats.java
@@ -1,6 +1,6 @@
 /* BSD 2-Clause (c) 2014: A.Doherty (Oxford), Shing Chan (Oxford)
  *
- * Code for extracting basic summary stats from raw triaxial 
+ * Code for extracting basic summary stats from raw triaxial
  * acceleration measurements.
  */
 import java.util.Arrays;
@@ -39,7 +39,7 @@ public class AccStats {
         }
         double[] enmoTrunc = trunc(enmo);
         double[] enmoAbs = abs(enmo);
-        
+
 
         // don't forget to change header method immediately below !!!
         double[] basicStatistics = {
@@ -49,17 +49,17 @@ public class AccStats {
             xStd, yStd, zStd,
             xyCovariance, xzCovariance, yzCovariance,
         };
-        
+
         double[] outputFeats = null;
         //extract features if requested
         if (getFeatures){
-            double[] features = Features.getFeatures(xArray, yArray, zArray, 
+            double[] features = Features.getFeatures(xArray, yArray, zArray,
                                             enmoTrunc, sampleRate, numFFTbins);
             outputFeats = AccStats.combineArrays(basicStatistics, features);
         } else{
             outputFeats = basicStatistics;
         }
-        
+
         return outputFeats;
     }
 
@@ -367,6 +367,6 @@ public class AccStats {
         }
         return output;
     }
-   
+
 
 }
--- a/java/AccStats.java
+++ b/java/AccStats.java
@@ -229,9 +229,9 @@ public class AccStats {
         // covariation
         double cov = sxy / n - sx * sy / n / n;
         // standard error of x
-        double sigmax = Math.sqrt(sxx / n -  sx * sx / n / n);
+        double sigmax = Math.sqrt(Math.max(0, sxx / n -  sx * sx / n / n));
         // standard error of y
-        double sigmay = Math.sqrt(syy / n -  sy * sy / n / n);
+        double sigmay = Math.sqrt(Math.max(0, syy / n -  sy * sy / n / n));
 
         // correlation is just a normalized covariation
         return cov / sigmax / sigmay;
--- a/java/Features.java
+++ b/java/Features.java
@@ -32,7 +32,7 @@ public class Features {
             int numFFTbinsPerChannel){
 
         // get San Diego (Ellis) features
-        double[] sanDiegoFeats = calculateSanDiegoFeatures(x, y, z, sampleRate, 
+        double[] sanDiegoFeats = calculateSanDiegoFeatures(x, y, z, sampleRate,
                                                             numFFTbinsPerChannel);
 
         // get MAD features
@@ -140,17 +140,17 @@ public class Features {
         double xyCorrelation = AccStats.correlation(wx, wy);
         double xzCorrelation = AccStats.correlation(wx, wz);
         double yzCorrelation = AccStats.correlation(wy, wz);
-        
+
         // Roll, Pitch, Yaw
         double [] angleAvgStdYZ = AccStats.angleAvgStd(wy, wz); //roll
         double [] angleAvgStdZX = AccStats.angleAvgStd(wz, wx); //pitch
         double [] angleAvgStdYX = AccStats.angleAvgStd(wy, wx); //yaw
-        
+
         // gravity component angles
         double gxyAngle = Math.atan2(gyMean,gzMean);
         double gzxAngle = Math.atan2(gzMean,gxMean);
         double gyxAngle = Math.atan2(gyMean,gxMean);
-        
+
         // don't forget to change header method immediately below !!!
         double[] output = new double[]{
             sdMean,
@@ -239,12 +239,12 @@ public class Features {
                 }
             }
         }
-        
+
         // column means
         double gxMean = AccStats.mean(gx);
         double gyMean = AccStats.mean(gy);
         double gzMean = AccStats.mean(gz);
-        
+
         return new double[] {gxMean, gyMean, gzMean};
     }
 
@@ -256,13 +256,13 @@ public class Features {
     {
         final int n = v.length;
         final double vMean = AccStats.mean(v);
-        
+
         // Initialize array to compute FFT coefs
         double[] vFFT = new double[n];
         for (int i = 0; i < n; i++){
             vFFT[i] = v[i] - vMean;  // note: we remove the 0Hz freq
         }
-        
+
         HanningWindow(vFFT, vFFT.length);
         new DoubleFFT_1D(vFFT.length).realForward(vFFT);  // FFT library computes coefs inplace
         final double[] vFFTpow = getFFTpower(vFFT);  // parse FFT coefs to obtain the powers
@@ -312,7 +312,7 @@ public class Features {
         for (int i = 0; i < numBins; i++){
             binnedFFT[i] = 0;
         }
-        
+
         final int windowOverlap = sampleRate / 2;  // 50% overlapping windows
         final int numWindows = n / windowOverlap - 1;
         double[] windowFFT = new double[sampleRate];
@@ -384,11 +384,11 @@ public class Features {
 
         // features from paper:
         // Mean amplitude deviation (MAD) describes the typical distance of data points about the mean
-        double MAD = 0; 
+        double MAD = 0;
         // Mean power deviation (MPD) describes the dispersion of data points about the mean
         double MPD = 0;
         // Skewness (skewR) describes the asymmetry of dispersion of data points about the mean
-        double skew = 0; 
+        double skew = 0;
         // Kurtosis (kurtR) describes the peakedness of the distribution of data points
         double kurt = 0;
         for (int i = 0; i < n; i++) {
@@ -403,7 +403,7 @@ public class Features {
         MPD /= Math.pow(N, 1.5);
         skew *= N / ((N-1)*(N-2));
         kurt = kurt * N*(N+1)/((N-1)*(N-2)*(N-3)*(N-4)) - 3*(N-1)*(N-1)/((N-2)*(N-3));
-        
+
         // don't forget to change header method immediately below !!!
         return new double[] {
             MAD,
@@ -463,7 +463,7 @@ public class Features {
         }
         return header;
     }
-    
+
 
     private static double[] getFFTmagnitude(double[] FFT) {
         return getFFTmagnitude(FFT, true);
@@ -484,7 +484,7 @@ public class Features {
     }
 
 
-    
+
     private static double[] getFFTpower(double[] FFT, boolean normalize) {
         /*
          * Get powers from FFT coefficients
@@ -553,7 +553,7 @@ public class Features {
         //Compute FFT and power spectrum density
         final int n = v.length;
         final double vMean = AccStats.mean(v);
-        
+
         // Initialize array to compute FFT coefs
         double[] vFFT = new double[n];
         for (int i = 0; i < n; i++)  vFFT[i] = v[i] - vMean;  // note: we remove the 0Hz freq
@@ -610,7 +610,7 @@ public class Features {
         String header = "f1,p1,f2,p2,f625,p625,totalPower";
         return header;
     }
-    
+
 
     /**
      * From paper:
--- a/java/Features.java
+++ b/java/Features.java
@@ -136,10 +136,10 @@ public class Features {
         double[] paQuartiles = AccStats.percentiles(v, new double[] {0, 0.25, 0.5, 0.75, 1});
 
         //correlations
-        double autoCorrelation = AccStats.correlation(v, v, sampleRate);
-        double xyCorrelation = AccStats.correlation(wx, wy);
-        double xzCorrelation = AccStats.correlation(wx, wz);
-        double yzCorrelation = AccStats.correlation(wy, wz);
+        double autoCorrelation = correlation(v, v, sampleRate);
+        double xyCorrelation = correlation(wx, wy);
+        double xzCorrelation = correlation(wx, wz);
+        double yzCorrelation = correlation(wy, wz);
 
         // Roll, Pitch, Yaw
         double [] angleAvgStdYZ = AccStats.angleAvgStd(wy, wz); //roll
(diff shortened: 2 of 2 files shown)


R-Walmsley:
I don't entirely understand the context of this commit - @chanshing could you outline the problem it's solving? In particular, is there any reason it's only an issue for sigmax and sigmay, but not for the covariance (line 230 - and obviously a very similar formula with many of the same inputs)?

chanshing:
> I don't entirely understand the context of this commit - @chanshing could you outline the problem it's solving? In particular, is there any reason it's only an issue for sigmax and sigmay, but not for the covariance (line 230 - and obviously a very similar formula with many of the same inputs)?

Thanks Rosemary. It was reported a few times before that the features corrxy, corrxz, corryz tended to give NaNs. I think a likely reason for this is line 232 `Math.sqrt(sxx / n -  sx * sx / n / n);` which might be negative. Similarly for line 234. Finally the last bit (return) ensures once more that we don't divide by zero.

chanshing:
Further clarification: 232 and 234 are trying to compute the sqrt of the variance. Sometimes the variance can be exactly zero, but due to computer precision it may happen that we get -0.000000000001. So that max(0, ...) is there to avoid this. Further, if both variances of x and y are zero, we cannot compute the correlation in 237, so it is set to 0 (which is the limit as the variances go to zero?)

R-Walmsley:
In what situation are the variances 0 or near 0? If the variance is 0, surely the covariance is undefined? (Both mathematically as 0/0, and if you think about what covariance or correlation captures)

I'm not sure how you can resolve it by thinking about the limit as they go to 0 in this case (and try to get an answer for 0/0 that way). Can't you get pairs of sequences of random variables with variance tending to 0 but correlation taking different values? 

So isn't nan the right answer in that case? (I appreciate that, for practical reasons, it may not be!)

chanshing:
> In what situation are the variances 0 or near 0? If the variance is 0, surely the covariance is undefined? (Both mathematically as 0/0, and if you think about what covariance or correlation captures)
> 
> I'm not sure how you can resolve it by thinking about the limit as they go to 0 in this case (and try to get an answer for 0/0 that way). Can't you get pairs of sequences of random variables with variance tending to 0 but correlation taking different values?
> 
> So isn't nan the right answer in that case? (I appreciate that, for practical reasons, it may not be!)

Thanks for this. In our case would be if the device is perfectly still during the epoch. Correct me if I'm wrong: Cov(X, a) = 0 for a constant. Is it true that Corr(X, a) = 0?

chanshing:
So according to this https://stats.stackexchange.com/questions/267152/correlation-with-a-constant#:~:text=2%20Answers&text=This%20means%20that%20if%20one,you%20are%20dividing%20by%20zero). R assigns NaN when it's the case :rofl:

chanshing:
After a bit more thought, I wonder if we can assume something mild. Let's first consider Var(X) -> 0 and Var(Y) -> 0 _and_ Y is normally distributed (which I don't think is that crazy in a 30s window at rest where machine noise probably dominates). Can we _then_ say that Corr(X, Y) -> 0? :thinking:

chanshing:
Found this example too https://stats.stackexchange.com/questions/46410/what-is-the-correlation-coefficient-between-two-zero-random-variables  I guess it simply cannot be defined satisfactorily.

chanshing:
Ok so according to this https://stats.stackexchange.com/q/267152/105780 R assigns NaN when it's the case.
Found this nice example too https://stats.stackexchange.com/questions/46410/what-is-the-correlation-coefficient-between-two-zero-random-variables showing that it can't be defined as it can converge to anything.
But after some thinking I wonder if we can still argue, for our _particular_ case, that the correlation is 0 when the device is static, because then it's pretty much the machine noise that dominates, and my impression is that these are independent.

angerhang:
> Ok so according to this https://stats.stackexchange.com/q/267152/105780 R assigns NaN when it's the case.
> Found this nice simple example too https://stats.stackexchange.com/questions/46410/what-is-the-correlation-coefficient-between-two-zero-random-variables showing that it can't be defined as it can converge to anything.
> 
> But after some thinking I wonder if we can still argue, for our _particular_ case, that the correlation is 0 when the device is static, because then it's pretty much the machine noise that dominates, and my impression is that these are independent. The NaNs predominantly occurs when the device is static and a lowpass filter is applied to help remove noise.

Thanks both for the helpful explanations. I share @R-Walmsley 's sentiment here if we depart from the statistical definition and the usual behavior that is not consistent with other software. 

I think perhaps the best thing is to make an optional argument that determines the behavior when the numerator is zero. @chanshing also has a point on it makes practical sense just to treat things being zero but I suspect that we might shoot ourselves in the foot one day when doing some very subtle statistical analysis.

Additionally, is it possible to follow up this PR with a few test cases to check its behavior :D

R-Walmsley:
I don't have a strong view- I think I need to look more into the context of this in the code in terms of what it's then used for, 

But do we need to include these periods? If the device is perfectly still, couldn't we treat it as if it were unrecorded time (because it seems so unlikely it's actually wear time)?

chanshing:
Thanks both for your inputs. This function is used to generate the features 
```
autocorr
corrxy
corrxz
corryz
```

Regarding nonwear, the current rule is that std(x), std(y), std(z) are all < 13mg continuously for 1 hour.

Not the best of tests but I quickly checked the first epoch labelled **sleep** of the clean capture24 dataset:
```
array([[ 0.9242714 , -0.15475608,  0.353567  ],
       [ 0.9242714 , -0.15193762,  0.353567  ],
       [ 0.9242714 , -0.1632115 ,  0.353567  ],
       ...,
       [ 0.9242714 , -0.15193762,  0.353567  ],
       [ 0.9242714 , -0.15193762,  0.353567  ],
       [ 0.9242714 , -0.15193762,  0.353567  ]], dtype=float32)
```

Using `np.unique` on the x-axis shows 2969 out of 3000 values (~99%) are 0.9242714. On the z-axis it's a bit "better": 2547 values are 0.353567. On the y-axis it's a bit more normal. For this I get `corrxy=-0.04`, `corrxz=0.006` and `corryz=0.06`. But I think it's pure numerical luck that we didn't get NaN for this epoch. I'm trying to see if there's an argument that can be made to avoid the NaN here.

I think a reason for seeing lots of constant values is the precision of the device, which is 1/256 g in case of Axivity, so there's only so much variation (256 per g to be precise) it can capture.

chanshing:
Would it be defensible to say that at rest the axes are independent? A device with infinite precision would just measure white noise.

R-Walmsley:
Would we expect sensor noise to be axis independent? I can at least imagine mechanisms for it where noise in all axes was correlated (e.g. thinking of an analogy imagining a giant accelerometer sat on a table outside, and little gusts of wind would affect the whole device).

chanshing:
> Would we expect sensor noise to be axis independent? I can at least imagine mechanisms for it where noise in all axes was correlated (e.g. thinking of an analogy imagining a giant accelerometer sat on a table outside, and little gusts of wind would affect the whole device).

The wind here would be a true signal though and not machine noise. I would imagine that most manufacturers make sure that their devices don't have weird correlated noise, but I'm not sure!

R-Walmsley:
What is machine noise though? (I really don't know enough about this) What if it's something to do with the electricity, which would affect all axes at once?

chanshing:
Borrowing from the grave of my engineering knowledge, usually, we assume that whatever continuous measurements we collect come with an isotropic noise f(t) + eps, eps~N(0, small_sigma*I) but granted this is done for convenience most of the time :rofl:

chanshing:
> What is machine noise though? (I really don't know enough about this) What if it's something to do with the electricity, which would affect all axes at once?

Yep it's possible that noise is perfectly correlated in all axes.

R-Walmsley:
I also think probably the debate about the true noise mechanism is more-or-less irrelevant to what we should do about it? 

I guess as @angerhang said our main worry should be the possible downstream consequences of the different right or wrong answers we could give? I don't have much of an answer on that, but guess 0 ("as if it came from independent random noise in all axes, whether or not it actually does")

chanshing:
> I also think probably the debate about the true noise mechanism is more-or-less irrelevant to what we should do about it?
> 
> I guess as @angerhang said our main worry should be the possible downstream consequences of the different right or wrong answers we could give? I don't have much of an answer on that, but guess 0 ("as if it came from independent random noise in all axes, whether or not it actually does")

The machine noise is for: If we can accept that the machine noise (what the machine outputs when perfectly still, without any external force, and with infinite precision) is independent in each axis, then the answer to our problem is simply 0. Now, this function is not used elsewhere but only for computing the features mentioned above. I think what we can do is put a clear comment around to clarify the intended use?

angerhang:
> > I also think probably the debate about the true noise mechanism is more-or-less irrelevant to what we should do about it?
> > I guess as @angerhang said our main worry should be the possible downstream consequences of the different right or wrong answers we could give? I don't have much of an answer on that, but guess 0 ("as if it came from independent random noise in all axes, whether or not it actually does")
> 
> The machine noise is for: If we can accept that the machine noise (what the machine outputs when perfectly still, without any external force, and with infinite precision) is independent in each axis, then the answer to our problem is simply 0. Now, this function is not used elsewhere but only for computing the features mentioned above. I think what we can do is put a clear comment around to clarify the intended use?

I like how we are trying to be rigorous here. I think what we are trying to answer here is whether setting correlation to be zero makes sense when one variable is constant. As @chanshing pointed out, we do need to explain more the intended use of the function here. Further, just intuitively during sleep,  if one variable is constant, does make it sense to set the correlation to zero. Is this the evidence for no correlation or the absence of correlation evidence? I probably would prefer the latter explanation.

As it stands, as I've said, wouldn't it be nice to provide an optional argument to let the user choose what's the best for their use? From a prediction perspective, it does make sense to for us default NAs to zero as the model doesn't care about if it is the true correlation but when doing inference, probably NAs would still be better :D

chanshing:
This has been a great discussion. From this I implemented a few changes:

1. On the off chance someone decided to use our `AccStats.correlation()` for something else, I reverted the changes (returning NaN) but keep the fixes in the lines computing the sqrt of the variance, where I make sure it is nonnegative.
2. But in Features.java where `AccStats.correlation()` is called to compute our signal features, when a NaN is returned I replace it with 0.0. That is, we shall assume the axes are independent at rest.

So we are imposing our bias on what correlation should be at rest, and do this closer to the calling point in the code, while leaving the backend code `AccStats.correlation()` behaving statistically consistent on the off chance someone uses it or simply visits the code and gets confused. What do you think of these changes?

BTW you'll notice that I rebased the commits and also had to commit first a 'remove trailing spaces' because my editor does this automatically. I also renamed the PR to something more useful

chanshing:
BTW something I learned today and caught my attention that I think is very releavant to this problem is https://en.wikipedia.org/wiki/Covariance#Definition

> but this equation is susceptible to catastrophic cancellation (see the section on numerical computation below).

The equation it is referring to is exactly the way we are computing the covariance, so I think a further improvement can be to apply a more numerically stable algorithm to compute the covariance.

java/AccStats.java
```diff
@@ -367,6 +367,6 @@ public static int countStuckVals(
         }
         return output;
     }
-   
+
 
 }
```
angerhang, line 372: missing end of line :D 
Good to add this in even though it is not part of your changes :D

angerhang:
> BTW something I learned today and caught my attention that I think is very releavant to this problem is https://en.wikipedia.org/wiki/Covariance#Definition
> 
> > but this equation is susceptible to catastrophic cancellation (see the section on numerical computation below).
> 
> The equation it is referring to is exactly the way we are computing the covariance at the moment, so I think a further improvement can be to apply a more numerically stable algorithm to compute the covariance.

Oh interesting, this is very subtle. Don't know how much it would influence our eventual result but it is an easy fix so I'd create an issue for this.

angerhang:
> This has been a great discussion. From this I implemented a few changes:
> 
> 1. On the off chance someone decided to use our `AccStats.correlation()` for something else, I reverted the changes (returning NaN) but keep the fixes in the lines computing the sqrt of the variance, where I make sure it is nonnegative.
> 2. But in Features.java where `AccStats.correlation()` is called to compute our signal features, when a NaN is returned I replace it with 0.0. That is, we shall assume the axes are independent at rest. In practice in the code I wrap the backend function `AccStats.correlation()` in a _private_ function of Features.java.
> 
> So we are imposing our bias on what correlation should be at rest, and do this closer to the calling point in the code, while leaving the backend code `AccStats.correlation()` behaving statistically consistent on the off chance someone uses it or simply visits the code and gets confused. What do you think of these changes?
> 
> BTW you'll notice that I rebased the commits and also had to commit first a 'remove trailing spaces' because my editor does this automatically. I also renamed the PR to something more useful

Thanks a lot the update. I think I am happy for its current shape. But just before we merge this PR, I would really love to push to add a few test cases to the `correlation` function in `AccStats` to to be sure and as a good habit to stick to a test-driven development process :D

java/AccStats.java
chanshing: I see. In fact my editor added this but I think I deleted it.

chanshing:
> > This has been a great discussion. From this I implemented a few changes:
> > 
> > 1. On the off chance someone decided to use our `AccStats.correlation()` for something else, I reverted the changes (returning NaN) but keep the fixes in the lines computing the sqrt of the variance, where I make sure it is nonnegative.
> > 2. But in Features.java where `AccStats.correlation()` is called to compute our signal features, when a NaN is returned I replace it with 0.0. That is, we shall assume the axes are independent at rest. In practice in the code I wrap the backend function `AccStats.correlation()` in a _private_ function of Features.java.
> > 
> > So we are imposing our bias on what correlation should be at rest, and do this closer to the calling point in the code, while leaving the backend code `AccStats.correlation()` behaving statistically consistent on the off chance someone uses it or simply visits the code and gets confused. What do you think of these changes?
> > BTW you'll notice that I rebased the commits and also had to commit first a 'remove trailing spaces' because my editor does this automatically. I also renamed the PR to something more useful
> 
> Thanks a lot the update. I think I am happy for its current shape. But just before we merge this PR, I would really love to push to add a few test cases to the `correlation` function in `AccStats` to to be sure and as a good habit to stick to a test-driven development process :D

Thanks Hang. What's the best practice here? Do I have to add a test together with the PR? How does it go from here?

R-Walmsley:
> This has been a great discussion. From this I implemented a few changes:
> 
> 1. On the off chance someone decided to use our `AccStats.correlation()` for something else, I reverted the changes (returning NaN) but keep the fixes in the lines computing the sqrt of the variance, where I make sure it is nonnegative.
> 2. But in Features.java where `AccStats.correlation()` is called to compute our signal features, when a NaN is returned I replace it with 0.0. That is, we shall assume the axes are independent at rest. In practice in the code I wrap the backend function `AccStats.correlation()` in a _private_ function of Features.java.
> 
> So we are imposing our bias on what correlation should be at rest, and do this closer to the calling point in the code, while leaving the backend code `AccStats.correlation()` behaving statistically consistent on the off chance someone uses it or simply visits the code and gets confused. What do you think of these changes?
> 
> BTW you'll notice that I rebased the commits and also had to commit first a 'remove trailing spaces' because my editor does this automatically. I also renamed the PR to something more useful

This sounds a sensible way round to do it! I've reviewed the code and didn't spot any issues.

R-Walmsley:
> BTW something I learned today and caught my attention that I think is very releavant to this problem is https://en.wikipedia.org/wiki/Covariance#Definition
> 
> > but this equation is susceptible to catastrophic cancellation (see the section on numerical computation below).
> 
> The equation it is referring to is exactly the way we are computing the covariance at the moment, so I think a further improvement can be to apply a more numerically stable algorithm to compute the covariance.

Thanks for sharing, interesting to think through this!

angerhang:
> > > This has been a great discussion. From this I implemented a few changes:
> > > 
> > > 1. On the off chance someone decided to use our `AccStats.correlation()` for something else, I reverted the changes (returning NaN) but keep the fixes in the lines computing the sqrt of the variance, where I make sure it is nonnegative.
> > > 2. But in Features.java where `AccStats.correlation()` is called to compute our signal features, when a NaN is returned I replace it with 0.0. That is, we shall assume the axes are independent at rest. In practice in the code I wrap the backend function `AccStats.correlation()` in a _private_ function of Features.java.
> > > 
> > > So we are imposing our bias on what correlation should be at rest, and do this closer to the calling point in the code, while leaving the backend code `AccStats.correlation()` behaving statistically consistent on the off chance someone uses it or simply visits the code and gets confused. What do you think of these changes?
> > > BTW you'll notice that I rebased the commits and also had to commit first a 'remove trailing spaces' because my editor does this automatically. I also renamed the PR to something more useful
> > 
> > 
> > Thanks a lot the update. I think I am happy for its current shape. But just before we merge this PR, I would really love to push to add a few test cases to the `correlation` function in `AccStats` to to be sure and as a good habit to stick to a test-driven development process :D
> 
> Thanks Hang. What's the best practice here? Do I have to add a test together with the PR? How does it go from here?

Maybe in this case, end-to-end testing for the private method probably wouldn't be necessary :D  It is good to have a test file like `Tests/java/AccStatsTest.java` that has a few test cases to check the situations that you described above and ofc the function's normal behavior .

https://github.com/activityMonitoring/biobankAccelerometerAnalysis/pull/68/files#diff-1f17871a482aae745b1af33a5ddc596b is perhaps is a good starting point in terms of the structure. You shouldn't need to change any of the workflow files at all as all the unit tests will be detected by default.

java/AccStats.java
angerhang: Usually we should have a newline char at the end of a file :D 

https://unix.stackexchange.com/questions/18743/whats-the-point-in-adding-a-new-line-to-the-end-of-a-file provides a bit more insights.

PR #59, opened 2020-07-14
make use of cutpoint based definitions explicit

R-Walmsley:
The aim of this to avoid unexpected overwriting behaviour when either MVPA or VPA appear in the ML-based classes.


aidendoherty:
Many thanks @R-Walmsley - Great work.
I've now merged in these changes.

PR #61, opened 2020-07-27
fix bug in get_interrupts, wrong index slicing

chanshing:
Fix a bug in `get_interrupts` where the index slicing is wrong.
Instead of `e.index[i:i+2]` it should be `e.index[i-1:i+1]`


chanshing:
@aidendoherty @R-Walmsley Happy to hear your comments!

PR #62, opened 2020-09-23
add autogeneration of all folders to avoid errors when using the default

R-Walmsley:
This adds autogeneration of all required folder names in accUtils.py, which avoids some errors the defaults generate. Often these folders will be empty (depending on the command options) but including them anyway is more consistent with the rest of the code.


chanshing:
Thanks Rosemary. This seems to be adding other unrelated bits according to the diff, like meanSDstr and meanCIstr. Perhaps these can go separately in another PR? Or if you don't want to submit these one thing you could do is to branch out from current master and apply the changes specific to what this PR aims to achieve.

R-Walmsley:
Thanks for pointing this out. This is due to some slightly strange orders of merging different versions of the same repo. I thought I'd fixed this issue, but obviously not. [The thing that makes it a bit weird is that, if you look at files changed, I think the only change is the correct one?]

Perhaps the easiest thing is if I just close this PR and submit a new one?

chanshing:
Yeah I think that would be best. Based on https://gist.github.com/Chaser324/ce0505fbed06b947d962 the easiest is you wanna first create a feature/bugfix/etc branch off the upstream master and work from there, so that the PRs can be more modular.
Your master is a few commits ahead so I think the following should work in creating a new branch off the upstream master:

```bash
git checkout -b fix-create-folders  # A new branch off your master
git fetch upstream  # Fetch this repo
git reset --hard upstream/master  # Make your new branch == this repo's master
```

PR #63, opened 2020-09-24
autogenerate folder names to avoid default errors

R-Walmsley:


R-Walmsley:
This adds autogeneration of all required folder names in accUtils.py, which avoids some errors the defaults generate. Often these folders will be empty (depending on the command options) but including them anyway is more consistent with the rest of the code.

accelerometer/accUtils.py
```diff
@@ -147,12 +147,19 @@ def writeStudyAccProcessCmds(accDir, outDir, cmdsFile='processCmds.txt',
     nonWearDir = os.path.join(outDir, 'nonWear')
     stationaryDir = os.path.join(outDir, 'stationary')
     logsDir = os.path.join(outDir, 'clusterLogs')
+    rawDir = os.path.join(outDir, 'raw') 
+    npyDir = os.path.join(outDir, 'npy') 
+    outputDir = os.path.join(outDir)
```
chanshing, line 152: This bit is not needed. `os.path.join` is for joining two path strings, e.g. `os.path.join('a/b', 'c')` will produce the string `'a/b/c'`. You get the same result by simply doing `'a/b' + '/' + 'c'`, but using `os.path.join` is safer because it handles trailing slahes `'/'` automatically (so `os.path.join('a/b/', 'c')` will also produce correctly `'a/b/c'`).

chanshing: So here you can simply remove this line, and use `outDir` directly below instead of `outputDir`

chanshing, line 152: This bit is not needed. os.path.join is for joining two path strings, e.g. `os.path.join('a/b', 'c')` will produce the string `'a/b/c'`. You get the same result with `'a/b' + '/' + 'c'`, but using os.path.join is safer because it handles trailing slahes `'/'` automatically (so `os.path.join('a/b/', 'c')` will also produce correctly `'a/b/c'`).

So here you can simply remove this line, and use `outDir` directly below instead of `outputDir`

chanshing:
Thanks Rosemary! Everything looks great. I just have a small comment (above)

R-Walmsley:
Should now be ready to go, but @chanshing could you confirm?

chanshing:
> Should now be ready to go, but @chanshing could you confirm?

I think 185 will also need to change to `outDir`?

R-Walmsley:
I *think* that's now done? Thanks for the hint, I have so many places I should probably do this !!

chanshing:
Sweet :dancers:  Everything looks great. @angerhang thoughts?

angerhang:
> Sweet 👯 Everything looks great. @angerhang thoughts?

Seems good to me :D

PR #64, opened 2020-09-28
Add Junit tests framework 

angerhang:
* The Junit testing workflow
* Sample unit test
* Timestamp conversion for gt3x v2
* Regression tests to parse gt3x v2 and checking the content against a template csv file

PR #65, opened 2020-09-30
add junit framework

angerhang:
Split the original PR https://github.com/activityMonitoring/biobankAccelerometerAnalysis/pull/64 into two:

This is the first one that introduces the testing framework.


angerhang:
Closing https://github.com/activityMonitoring/biobankAccelerometerAnalysis/pull/67 as it is being reinstated using a pr from frok.

PR #66, opened 2020-09-30
Fix gt3x v2

angerhang:
The second PR for https://github.com/activityMonitoring/biobankAccelerometerAnalysis/pull/64.
This one needs to be merged before https://github.com/activityMonitoring/biobankAccelerometerAnalysis/pull/65 is merged.

Changes included:
1. A regression test to evaluate if the gt3x v2 parsing is successful and whether the epoch file content has changed.
2. It confirms the time extraction for actigraph. The current implementation still requires post-processing to ensure the timing correctness. 

What needs to be done in the future:
For gt3x V2, the time is being stored in .NET tick which describes the time difference in nanoseconds from 0001-01-01. So to get the local time, we will first need to extract 1970-01-01 in nanoseconds from the time tick. Then we need to move by the time shift to obtain `UNIX` time then one can use the `UNIX` time as one wishes.

However, we don't always know what time zone gt3x v2 has. So by directly converting the time ticks w.r.t UNIX time, most of the timestamps will be fine except when the cross-over happens because different timezones change at different relative times. 

So for future work, we could perhaps allow the user to supply an optional time zone argument for gt3x v2, otherwise, we should assume the time zone to be UTC. Currently, the default is `Europe/London`. 

For the time being, it is probably fine to require the user to post-process the date to get the right time. We shall create an issue after this PR is merged.


angerhang:
closing as changes are reinstated in https://github.com/activityMonitoring/biobankAccelerometerAnalysis/pull/68.

PR #67, opened 2020-10-01
add junit framework

angerhang:
Introducing the CI testing framework.

* Added workflow files 
* Provided a few sample tests


chanshing:
Thanks Hang. Everything looks fine. A bit doubtful about having to convert a method to public for the sole purpose of testing (what if we had to test _all_ methods?). Found this https://www.geeksforgeeks.org/how-to-call-private-method-from-another-class-in-java-with-help-of-reflection-api/ but I guess since you're dealing with some other things we might as well accept this :man_dancing: . What do you think @aidendoherty ? :dancers:

chanshing:
Thanks Hang. Everything looks fine. A bit doubtful about having to convert a method to public for the sole purpose of testing (what if we had to test _all_ methods?). Found this https://www.geeksforgeeks.org/how-to-call-private-method-from-another-class-in-java-with-help-of-reflection-api/ but I guess since you're dealing with many other things I think this is good to go :man_dancing: . What do you think @aidendoherty @R-Walmsley  ? :dancers:

angerhang:
> Thanks Hang. Everything looks fine. A bit doubtful about having to convert a method to public for the sole purpose of testing (what if we had to test _all_ methods?). Found this https://www.geeksforgeeks.org/how-to-call-private-method-from-another-class-in-java-with-help-of-reflection-api/ but I guess since you're dealing with many other things I think this is good to go 🕺 . What do you think @aidendoherty @R-Walmsley ? 👯

Thanks for the comment. Actually this PR was meant to be only concerns CI and nothing else. So Removed the gt3x testing to the other PR. https://github.com/activityMonitoring/biobankAccelerometerAnalysis/pull/68

chanshing:
I see. I think I will just go ahead and merge it since Rosemary already reviewed it in the previous PR :cow:

PR #70, opened 2020-10-05
Fixed gt3x v1 and v2 parsing with ci tests

angerhang:
Okay, after much struggle with the timeshift for v1 and v2, we are now finally working with a clean version with the following changes:

1. timestamp changes for gt3x v1 and v2
2. Junit tests for timeshift correction
3. A regression test to evaluate if the gt3x parsing is successful and whether the epoch file content has changed.

Basically the changes are based on the fact that the timestamps in the `info.txt` are in [.NET ticks](https://docs.microsoft.com/en-us/dotnet/api/system.datetime.ticks?redirectedfrom=MSDN&view=netcore-3.1), and the timestamps for the activities are stored in UNIX time style but is actually local time.

So to correctly parse the time, we need to add the timeshift to the activity data timestamps and then convert them to the right time zones. Since we don't know which time zones precisely from the files themselves, so a user will need to manually adjust that during the post-processing.


java/ActigraphReader.java
```diff
@@ -72,31 +78,41 @@ public static void readG3TXEpochs(
 
             // underscored are unused for now
             double sampleFreq = -1, accelerationScale = -1, _AccelerationMin, _AccelerationMax;
-            long _LastSampleTime, firstSampleTime=-1;
+            long startDate = -1, stopDate = -1, firstSampleTime=-1;
             String serialNumber = "";
+            String infoTimeZone = "01:00:00"; // default to be UTC time difference
```
chanshing, line 83: Shouldn't this be something like `UTC` or `Europe/London` instead of some time offset? A time offset may correspond to many timezones. Also even if it's UTC I guess it should be `"00:00:00"` instead?

java/ActigraphReader.java
```diff
@@ -72,31 +78,41 @@ public static void readG3TXEpochs(
...
                             serialNumber=tokens[1].trim();
+                        else if (tokens[0].trim().equals("TimeZone"))
+                            infoTimeZone=tokens[1].trim();
                     }
                 }
             }
+
+            System.out.println("Timezone: " + infoTimeZone);
+            System.out.println("Start date (local UNIX): " + startDate);
+            System.out.println("Stop date (local UNIX): " + stopDate);
```
chanshing, line 114: This will print the pseudo/local UNIX times, correct? Is this here just for debugging?

java/ActigraphReader.java
```diff
@@ -147,6 +165,7 @@ else if (tokens[0].trim().equals("Serial Number"))
      **/
     private static void readG3TXV2Epoch(
             InputStream activityReader,
+            String infoTimeShift,
```
chanshing, line 168: Aha! So it _is_ a time shift after all, not timezone

chanshing:
Nice work Hang. Obviously I haven't put as much effort as you into this and haven't worked much with gt3x device, so my review will be limited. I could only point out something strange with the variable name where I think you meant time shift instead of timezone (see review). Hope this helps.

chanshing (changes requested):

java/ActigraphReader.java
angerhang: For `gt3x` files, they call the time shift as timezone because we don't have the timezone information.

java/ActigraphReader.java
angerhang: Not just for debugging but to print out what's being extracted from the meta file.

java/ActigraphReader.java
angerhang: yeah

angerhang:
> Nice work Hang. Obviously I haven't put as much effort as you into this and haven't worked much with gt3x device, so my review will be limited. I could only point out something strange with the variable name where I think you meant time shift instead of timezone (see review). Hope this helps.

Thanks for the feedback. I know it is confusing but I tried to keep all the variable names consistent this time :D

java/ActigraphReader.java
```diff
@@ -80,7 +80,7 @@ public static void readG3TXEpochs(
             double sampleFreq = -1, accelerationScale = -1, _AccelerationMin, _AccelerationMax;
             long startDate = -1, stopDate = -1, firstSampleTime=-1;
             String serialNumber = "";
-            String infoTimeZone = "01:00:00"; // default to be UTC time difference
+            String infoTimeShift = "01:00:00"; // default to be UTC time difference
```
chanshing, line 83: For UTC shouldn't this be `"00:00:00"`?

java/ActigraphReader.java
```diff
@@ -104,12 +104,12 @@ else if (tokens[0].trim().equals("Stop Date"))
...
                             serialNumber=tokens[1].trim();
                         else if (tokens[0].trim().equals("TimeZone"))
-                            infoTimeZone=tokens[1].trim();
+                            infoTimeShift=tokens[1].trim(); // gt3x calls time shift as time zone
                     }
                 }
             }
 
-            System.out.println("Timezone: " + infoTimeZone);
+            System.out.println("Timezone: " + infoTimeShift);
```
chanshing, line 112: What do you think about making this more explicit, like "Time shift wrt. local" or just "Time shift" instead of "Timezone"? I'm thinking in a few years time when our memories have faded it would be less confusing for future developers/phd students/postdocs :student:

chanshing (changes requested):

java/ActigraphReader.java
```diff
@@ -104,12 +104,12 @@ else if (tokens[0].trim().equals("Stop Date"))
                         else if (tokens[0].trim().equals("Serial Number"))
                             serialNumber=tokens[1].trim();
                         else if (tokens[0].trim().equals("TimeZone"))
-                            infoTimeZone=tokens[1].trim();
+                            infoTimeShift=tokens[1].trim(); // gt3x calls time shift as time zone
```
chanshing, line 107: Great clarification

java/ActigraphReader.java
angerhang: Hmm this is really just printing out the TimeZone field from `info.txt` because that's how it is written. So not sure if we should change this.

java/ActigraphReader.java
angerhang: I think you are right. Let me test it out.

R-Walmsley:
Looks good to me (although I'm not super familiar with this end of the code). One quick question I had was where the sample data comes from?

angerhang:
> Looks good to me (although I'm not super familiar with this end of the code). One quick question I had was where the sample data comes from?

Good question. It was actually taken from https://github.com/THLfi/read.gt3x. They have an EUPL licence. Not sure if that's going to cause any issues? Having read our license, I am not sure what kind of license we have which makes it difficult to understand if it is compliant with EUPL.

java/ActigraphReader.java
chanshing: Personally would prefer not to perpetuate the confusing manufacturer's decision making. Since this print out is exposed to the end user I think it would be better to just print time shift. @R-Walmsley @aidendoherty what do you think about this? Context: GT3X's internal bookkeeping has a variable 'TimeZone' but it is actually a time shift wrt. the local time.

java/ActigraphReader.java
angerhang: Yeah `00:00:00` is actually correct after I've tested them out on different files :D Good catch!

R-Walmsley:
> > Looks good to me (although I'm not super familiar with this end of the code). One quick question I had was where the sample data comes from?
> 
> Good question. It was actually taken from https://github.com/THLfi/read.gt3x. They have a EUPL licence. Not sure if that's going to cause any issues? Having read our license, I am not sure what kind of license we have which makes it difficult to understand if it is compliant with EUPL.
 

It's a[ BSD license](https://en.wikipedia.org/wiki/BSD_licenses) I think? Which based on [this](https://joinup.ec.europa.eu/collection/eupl/licence-compatibility) would need to become a EUPL license if we include this code?

java/ActigraphReader.java
chanshing: Come to think of it, there's also a `--timeZone` flag (default Europe/London) in accProcess.py which would add to the confusion. Perhaps a compromise would be to print "Device timezone" but again I'd prefer "Device timeshift"

angerhang:
> > > Looks good to me (although I'm not super familiar with this end of the code). One quick question I had was where the sample data comes from?
> > 
> > 
> > Good question. It was actually taken from https://github.com/THLfi/read.gt3x. They have a EUPL licence. Not sure if that's going to cause any issues? Having read our license, I am not sure what kind of license we have which makes it difficult to understand if it is compliant with EUPL.
> 
> It's a[ BSD license](https://en.wikipedia.org/wiki/BSD_licenses) I think? Which based on [this](https://joinup.ec.europa.eu/collection/eupl/licence-compatibility) would need to become a EUPL license if we include this data?

Rosemary, you are right. It took me a while to find the data that we could use. This is going to be an issue in general for all of our end-to-end pipelines if we can publicly use the data. @chanshing  @aidendoherty any suggestions on this?

For this case, I managed to find an alternative under GNU this time. I would make the appropriate changes to be GNU compliant. I mean one could even go as far as our license could be incompatible with all the packages we use because Python different packages actually could have different licenses :(

java/ActigraphReader.java
angerhang: Yeah maybe it is confusing. From a user perspective actually, it would be odd to change the name of the extracted field because all else have the same field names except for the time zone.  Other software packages also used time zone https://github.com/THLfi/read.gt3x/blob/c4b83cc262c7b3dc66b0408f9c72123686dd3e33/R/readGT3X.R#L187

Maybe just to avoid confusion with two time zones being the same thing, we shall use `device timezone`.

chanshing: It's so confusing in more than one way :sweat_smile: I read this [opinionated blog](https://spin.atomicobject.com/2016/07/06/time-zones-offsets/#:~:text=A%20time%20zone%20is%20a,because%20of%20Daylight%20Saving%20Time.) some time ago when I was fixing the Axivity timestamps, and read it again today. Timezones are `"Europe/London"`, `"GMT"`, etc. and time _offsets_ are `"+01:00"`, `"-09:00"`, etc. but people usually mix them up. They shouldn't because a single timezone can have multiple offsets (like Europe/London which is +01 in Apr-Oct and +00 in Oct-Apr).

Adding to the confusion is the fact that, as you know, the device's offset is really only accurate at instant t=0 :sweat_smile: because any time after that can suffer a crossover and now we are in a different offset. That's why I wasn't tempted to call it _offset_ (nor timezone), just "timeshift". To be absolutely strict one would have to print the more verbose `"Device's initial offset"`

I don't think it would be that odd to change the name, as we already do inside the code. Are users likely to examine info.txt?

java/ActigraphReader.java
```diff
@@ -443,7 +467,22 @@ private static void readG3TXV1EpochPairs(
...
+    public static long getTrueUnixTime(long myTime, String infoTimeShift) {
+        int shiftSign = 1;
+        if (infoTimeShift.charAt(0) == '-') {
+            shiftSign = -1;
+            infoTimeShift = infoTimeShift.substring(1);
+        }
+
+        LocalTime timeShift = LocalTime.parse(infoTimeShift);
+        long timeShiftMilli = 1000 * (shiftSign * timeShift.getHour() * 60 * 60 +
+                timeShift.getMinute() * 60 + timeShift.getMinute()); // time shfit w.r.t. UTC
```
chanshing, line 479: Related to your issue discussed in Teams, can you confirm these lines are OK? I'm not at all familiar with this device but it would seem that the minute is added twice?

angerhang: Think you are right. This is a mistake. Thanks for catching this.

angerhang:
> > > Looks good to me (although I'm not super familiar with this end of the code). One quick question I had was where the sample data comes from?
> > 
> > 
> > Good question. It was actually taken from https://github.com/THLfi/read.gt3x. They have a EUPL licence. Not sure if that's going to cause any issues? Having read our license, I am not sure what kind of license we have which makes it difficult to understand if it is compliant with EUPL.
> 
> It's a[ BSD license](https://en.wikipedia.org/wiki/BSD_licenses) I think? Which based on [this](https://joinup.ec.europa.eu/collection/eupl/licence-compatibility) would need to become a EUPL license if we include this data?

Now the gt3x file is taken from https://github.com/muschellij2/gt3x_comparison where no license is stated. So I think we could just use it in our repo.

angerhang:
V1 files have also been validated using against  https://github.com/actigraph/NHANES-GT3X-File-Format/releases that @shing kindly helps to process. Think it is ready to go.

java/ActigraphReader.java
chanshing: Hi Hang, what's your final opinion on this? My preference is that we print `Device's initial offset` instead of `Timezone` (which might be confused with the user-inputted timezone, e.g. `Europe/London`.

chanshing (changes requested):
Everything looks great and thanks for the fixes and tests. Just a final comment on terminology (comments below). Perhaps @R-Walmsley and @aidendoherty can weigh in as well?

angerhang:
> Everything looks great and thanks for the fixes and tests. Just a final comment on terminology (comments below). Perhaps @R-Walmsley and @aidendoherty can weigh in as well?

Makes sense. People might confuse this with the user argument. Just changed.

chanshing:
Everything looks great. IMO it's ready to go.

Issue #73, opened 2020-10-06
Systematic makeover of time zone 


angerhang:
So I am trying to process a gt3x file with a flag of `--timeZone Australia/Perth`. The extracted epoch gives me the right timestamp. `2016-04-12 18:00:00.033+0800 [Australia/Perth]` is the time for the first epoch. However, in summary, this is what I obtained:

```json
2020-10-06 22:58:28	=== Short summary ===
{
    "file-startTime": "2016-01-18 10:00:00",
    "file-endTime": "2016-01-25 09:59:30",
    "acc-overall-avg": 26.71094,
    "wearTime-overall(days)": 6.96,
    "nonWearTime-overall(days)": 0.04,
    "quality-goodWearTime": 1
}
```

All the above are shown in UTC in fact. The above output is in `UTC` time but the ideal behaviour should really be `Australia/Perth` time. Moreover, the time zone is not shown here, so some careless users might mistake the summary time being the local time too. We probably should make sure that our generated analytics all have the time zone information where possible. :> 

This might touch upon many different places, we probably should discuss what's the best way to resolve this issue.

@chanshing @aidendoherty @R-Walmsley

chanshing:
Another great catch @angerhang. Thanks a lot for all the effort in testing the code.
Yup I think we (I) forgot to update the write-out of the summary json file to reflect the recent major changes to the handling of timestamps. So I guess to make it consistent it'd be good to print out with the same format used in the epoch files. In accUtils.py there's the function `date_strftime` which I believe you can use to this end.

Issue #71, opened 2020-10-06
tz-naive and tz-aware date time comparison


angerhang:
While I was using the `--endTime 2017-04-18T18:00` in a format that is compliant with the documentation https://github.com/activityMonitoring/biobankAccelerometerAnalysis/blob/0efaf0267d81121b0e244979c3efc1635d7fc8c2/accProcess.py#L51-L52 I ran to a type error 
```python
TypeError: Cannot compare tz-naive and tz-aware datetime-like objects
```
as it was trying to compare two timezone-naive and timezone-aware objects together:

This was caused by https://github.com/activityMonitoring/biobankAccelerometerAnalysis/blob/4ffefac5861a1ec58c28a8fe2bc3bee249cb93e2/accelerometer/summariseEpoch.py#L92-L93

as `e.index` is tz-aware but `endTime` is not. Probably we will have the same issue, when we supply the start time option.

I think https://github.com/activityMonitoring/biobankAccelerometerAnalysis/commit/379ebb2025f2ed44334c631288b85b2f3e660e6b  introduced the error as our output became time zone aware. 

There are two ways that we could resolve this issue, one is to change the `endTime` and `startTime` type and make them tz-aware. Another way is to convert `e.index` to be tz-naive for comparison. But the former approach is probably better. 

@chanshing what do you think, since you worked on this piece of code? :D

chanshing:
Thanks for identifying this @angerhang 
I think the first approach will involve less work, i.e. to localize the `endTime`.
You could try to momentarily convert the index to tz-naive but then you'd need to make sure everything is right.

angerhang:
Fixed in pr #74.

PR #74, opened 2020-10-08
Fix startTime and endTime comparison in summariseEpoch

angerhang:
Fix the problem in https://github.com/activityMonitoring/biobankAccelerometerAnalysis/issues/71 and https://github.com/activityMonitoring/biobankAccelerometerAnalysis/issues/73

### DIFF
--- a/accelerometer/summariseEpoch.py
+++ b/accelerometer/summariseEpoch.py
@@ -87,10 +87,13 @@ def getActivitySummary(epochFile, nonWearFile, summary,
 
     # Remove data before/after user specified start/end times
     rows = e.shape[0]
+    tz = pytz.timezone(timeZone)
     if startTime:
-        e = e[e.index >= startTime]
+        localStartTime = tz.localize(startTime)
+        e = e[e.index >= localStartTime]
     if endTime:
-        e = e[e.index <= endTime]
+        localEndTime = tz.localize(endTime)
+        e = e[e.index <= localEndTime]
     # Quit if no data left
     if e.shape[0] == 0:
         print("No rows remaining after start/end time removal")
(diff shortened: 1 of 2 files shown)


accelerometer/summariseEpoch.py
```diff
@@ -87,21 +87,24 @@ def getActivitySummary(epochFile, nonWearFile, summary,
...
         print("Previously there were %d rows, now shape: %s" % (rows, str(e.shape)))
         sys.exit(-9)
 
     # Get start & end times
-    startTime = pd.to_datetime(e.index.values[0])
-    endTime = pd.to_datetime(e.index.values[-1])
-    summary['file-startTime'] = startTime.strftime('%Y-%m-%d %H:%M:%S')
-    summary['file-endTime'] = endTime.strftime('%Y-%m-%d %H:%M:%S')
+    startTime = pd.to_datetime(e.index.values[0]).tz_convert(tz)
+    endTime = pd.to_datetime(e.index.values[-1]).tz_convert(tz)
```
chanshing, line 105: Is it needed to convert `e.index` again? I think it's already in `tz` time...

chanshing (commented):
Great work!

accelerometer/summariseEpoch.py
angerhang: Thanks for spotting this. Fixed it. Well, I can't show the test result yet since the gt3x test cases haven't been merged yet.

R-Walmsley:
Can I confirm I understand what's changing here? 

- startTime and endTime are specified in local time. 
- Then they are converted to tz-aware with the time zone used throughout. 
- They are then compared with e.index, which is also now tz-aware?

angerhang:
> * They are then compared with e.index, which is also now tz-aware?

exactly :D

R-Walmsley (commented):
Looks good!

chanshing:
Shall we merge this or is this waiting for the gt3x PR? I think it is independent but not sure @angerhang

angerhang:
> Shall we merge this or is this waiting for the gt3x PR? I think it is independent but not sure @angerhang

Think is independent :D

PR #76, opened 2020-10-15
fix start/endTime flag time zone

angerhang:
Even though https://github.com/activityMonitoring/biobankAccelerometerAnalysis/pull/74 allows us to compare endTime and startTime properly. Both of these values had a default time zone of `UTC` and doesn't change with user argument.  This PR fixes the issue by giving start/endTime the right time zone info.

### DIFF
--- a/java/AccelerometerParser.java
+++ b/java/AccelerometerParser.java
@@ -69,6 +69,8 @@ public class AccelerometerParser {
     	boolean useFilter = true;
     	long startTime = -1; // milliseconds since epoch
     	long endTime = -1;
+    	String startTimeStr = "";
+    	String endTimeStr = "";
     	boolean getFeatures = false;
     	int numFFTbins = 12; // number of fft bins to print
         // Must supply additional information when loading from a .csv file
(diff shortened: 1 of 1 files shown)


java/AccelerometerParser.java
```diff
@@ -187,7 +180,23 @@ public static void main(String[] args) {
...
+					System.exit(-2);
+				}
+			}
+			if (!endTimeStr.isEmpty()) {
+				try {
+					endTime = dateFormat.parse(endTimeStr).getTime();
+				} catch (ParseException ex) {
+					System.err.println("error parsing endTime:'" + endTimeStr + "', must be in format: 1996-7-30T13:59");
+					System.exit(-2);
+				}
```
chanshing, line 199: Any particular reason why this is moved to the bottom?

chanshing (approved):
Great catch and thanks for fixing it Hang!

java/AccelerometerParser.java
angerhang: I know. The main reason for this is that we can't guarantee that lateTime or startTime flag will be read after the timeZone flag. 

That's why we should only do the conversion after all the flags have been read.

R-Walmsley (approved):
Looks good as far as I can tell!

angerhang:
Given this PR has been approved. I am just going to shamelessly merge this one myself :D

PR #77, opened 2020-10-16
add clipsBefore/AfterCalibr and total reads to the summary

R-Walmsley:
This adds three outputs relating to data quality to summary.json.

### DIFF
--- a/.github/workflows/data_quality_outputs.yml
+++ b/.github/workflows/data_quality_outputs.yml
@@ -24,9 +24,11 @@ jobs:
         run: bash utilities/downloadDataModels.sh
        
       - name: Summarise sample
-        run: import accelerometer
-             python accelerometer/summariseEpoch.py data/sample.cwa
+        run: |
+            import accelerometer
+            python accelerometer/summariseEpoch.py data/sample.cwa
           
       - name: compare output
-        run:  import accelerometer
-              python accelerometer/tests/check_data_quality_outputs.py
+        run:  |
+            import accelerometer
+            python accelerometer/tests/check_data_quality_outputs.py
--- a/.github/workflows/data_quality_outputs.yml
+++ b/.github/workflows/data_quality_outputs.yml
@@ -27,4 +27,4 @@ jobs:
         run: python accelerometer/summariseEpoch.py data/sample.cwa
           
       - name: compare output
-        run: python acceleromter/tests/check_data_quality_outputs.py
+        run: python accelerometer/tests/check_data_quality_outputs.py
--- a/.github/workflows/data_quality_outputs.yml
+++ b/.github/workflows/data_quality_outputs.yml
@@ -1,0 +1,32 @@
+#
+#  1. Check we get expected outputs from data quality parameters on sample data
+#
+name: data_quality_outputs
+
+on:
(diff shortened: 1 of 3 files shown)


angerhang (commented):
Hi Rosemary, thanks for the PR. Everything looks great. Do you think it is possible to mimic a test case as part of your PR just to make sure that we get the expected output?  
https://github.com/activityMonitoring/biobankAccelerometerAnalysis/pull/70/files

Of course, if you do need this PR to merged, then don't worry about this. Otherwise, I think it would be a good exercise.

R-Walmsley:
Good idea - do you mean as a test in the testing framework? (Which seems like the best plan, but I might need some guidance on how to do this, especially as I think the tests so far are all of the java parts?)

angerhang:
> Good idea - do you mean as a test in the testing framework? (Which seems like the best plan, but I might need some guidance on how to do this, especially as I think the tests so far are all of the java parts?)

Great! It is best to mimic  `.github/workflows/gt3x.yml` in https://github.com/activityMonitoring/biobankAccelerometerAnalysis/pull/70/files which compares the output of a parsed GT3X with an expected one. It hasn't been merged yet as it has the issue of the data file, but you could well use the `cwa` file that we already have :D 

This is a good start to know how to write github actions. Let me know if you have any other questions!
https://docs.github.com/en/free-pro-team@latest/actions

chanshing:
Everything looks great :) Feel free to merge if urgent @R-Walmsley @angerhang

chanshing (approved):

.github/workflows/data_quality_outputs.yml
```diff
@@ -0,0 +1,34 @@
...
+      
+      - name: Load data
+        run: bash utilities/downloadDataModels.sh
+       
+      - name: Summarise sample
+        run: |
+            import accelerometer
+            python accelerometer/summariseEpoch.py data/sample.cwa
+          
+      - name: compare output
```
angerhang, line 31: Actually, for the comparison, you can just save a sample summarize epcoh file and then using the below snippet to compare that with the generated one 

```python
      - name: compare epoch output
         uses: tianhaoz95/mirror-action@v1.0.1
         with:
           original: PATH2ORIGINAL
           mirror: PATH2GENERATED
```

.github/workflows/data_quality_outputs.yml
```diff
@@ -0,0 +1,34 @@
...
+      - name: Compile Parser
+        run: javac -cp java/JTransforms-3.1-with-dependencies.jar java/*.java
+        
+      
+      - name: Load data
+        run: bash utilities/downloadDataModels.sh
+       
+      - name: Summarise sample
+        run: |
+            import accelerometer
```
angerhang, line 28: Don't think you need to do `import accelerometer` but setting up the Python environment will do because the daemon running the tests will not have python or anything packages that we use by default. https://docs.github.com/en/free-pro-team@latest/actions/guides/building-and-testing-python could be of help :D

angerhang (changes requested):

R-Walmsley:
Closed to submit separate PRs for data quality outputs and CI.

PR #82, opened 2020-10-28
add clipsBefore/AfterCalibr and total reads to the summary.

R-Walmsley:
This only adds the extra outputs. Tests will be added in a separate PR (separating the elements of PR #77 ).

### DIFF
--- a/accelerometer/summariseEpoch.py
+++ b/accelerometer/summariseEpoch.py
@@ -117,6 +117,11 @@ def getActivitySummary(epochFile, nonWearFile, summary,
     get_wear_time_stats(e, epochPeriod, stationaryStd, minNonWearDuration,
         nonWearFile, summary)
 
+    # Calculate and include data quality statistics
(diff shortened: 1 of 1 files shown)


angerhang:
Thx for the great work Rosemary. Hmm I am just a bit confused. I think it would have been OK for the tests and the features changes to land in the same PR otherwise how would we know the changes being made are for certain correct except for eyeballing them? :D

R-Walmsley:
We're not yet adding unit tests for those features, we're adding an overall regression test for the output? Of which this is only 1 tiny aspect, so in my head it made sense to separate them... Happy to be overruled on this though!

R-Walmsley:
Actually maybe the best thing is to leave this unmerged until we've merged the regression test, then we will check it? 
I'm just having more trouble with the test!

R-Walmsley:
@angerhang  - I'd like to merge this then immediately merge the regression test on top of this? Is that ok with you?

PR #83, opened 2020-10-28
Add regression testing

R-Walmsley:
Unfinished as failing but submitting so @angerhang can suggest changes.


.github/workflows/regression_test_summariseEpoch.yml
```diff
@@ -0,0 +1,43 @@
+name: Python - summariseEpoch.py
+
+on: [push]
```
angerhang, line 3: Change this to
```python
on:
  push:
    branches: [ master ]
  pull_request:
    branches: [ master ]
```
so that this workflow is also run when a PR is being requested.

.github/workflows/regression_test_summariseEpoch.yml
```diff
@@ -0,0 +1,43 @@
...
+
+jobs:
+  build:
```
angerhang, line 6: the name of this job shouldn't be `build` because it is trying to check the summary. Maybe something `summary-check` would be better?

.github/workflows/regression_test_summariseEpoch.yml
```diff
@@ -0,0 +1,43 @@
...
+   #     # stop the build if there are Python syntax errors or undefined names
+   #     flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
+   #     # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
+   #     flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
+    - name: Compile Parser
+      run: javac -cp java/JTransforms-3.1-with-dependencies.jar java/*.java
+    - name: Load data
+      run: bash utilities/downloadDataModels.sh
+    - name: Process data using summariseEpoch.py
+      run: python accelerometer/summariseEpoch.py data/sample.cwa
```
angerhang, line 35: I am very confused here. Shouldn't we have the epoch as input instead of ``sample.cwa`` ? 

Also I guess what we really want to do here is to parse the `sample.cwa` by `python3 accProcess.py data/sample.cwa.gz`? Then compare the output epoch and summary files?

.github/workflows/regression_test_summariseEpoch.yml
```diff
@@ -0,0 +1,43 @@
...
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install flake8 pytest
+        if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
+   # - name: Lint with flake8
+   #   run: |
```
angerhang, line 25: This block can be deleted if we don't need them.

accelerometer/tests/check_data_quality_outputs.py
```diff
@@ -0,0 +1,13 @@
+import json
```
angerhang, line 1: Also where are we calling this script to check the output? Maybe it would be easier if we also have a reference json file instead and use the `tianhaoz95/mirror-action@v1.0.1` for comparison?

angerhang (changes requested):
See the comments.

.github/workflows/regression_test_summariseEpoch.yml
```diff
@@ -0,0 +1,43 @@
+name: Python - summariseEpoch.py
```
angerhang, line 1: Perhaps also good to change the name of this workflow into something more meaningful. See https://github.com/activityMonitoring/biobankAccelerometerAnalysis/pull/84 here for what those checks look like.

.github/workflows/regression_test_summariseEpoch.yml
R-Walmsley: yes it was meant to be accProcess.py, oops! Thanks for spotting this.

.github/workflows/regression_test_summariseEpoch.yml
R-Walmsley: Think it would be a good idea to add the linting at some point, once a basic version is working (it's just a bit of work to get the right settings and to run it the first time, so wanted to sort everything else first).

accelerometer/tests/check_data_quality_outputs.py
R-Walmsley: This is meant to have disappeared, will delete!

R-Walmsley:
Currently pushing just so @angerhang can help diagnose problems and review code - please don't merge it until I've squashed all the commits :)

angerhang (approved):

R-Walmsley:
@angerhang I'd like to squash these commits - can I still do that?

angerhang:
Oh crap sorry. I forgot that needed to be done. Hmm yes but then you will have to do it on the master which is a bit dangerous.

PR #85, opened 2020-11-03
Add regression test; Add data quality outputs in summary

R-Walmsley:
This should replace #83 as a single clean commit. To actually achieve this, I think we need to somehow rebase master to pick out the messy series of commits in the merge from #83, and then merge this. However, this is obviously a little risky and needs to be done carefully. 

@chanshing @angerhang can you confirm that you agree this is a reasonable plan? 

In the future, can we not merge before checking whether it would be good to squash the commits? I was adding lots of small commits to #83 to try to fix it (because I had to run the Actions and collaborate on it), and was always planning to squash them before merging... Obviously it's not crucial from a technical point of view, but the commit history will be harder to read for future users if we let this sort of thing remain in it.

### DIFF
--- a/.github/workflows/regression_test_summary.yml
+++ b/.github/workflows/regression_test_summary.yml
@@ -1,0 +1,48 @@
+name: regression test for processing sample data to epoch file
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  summary-check:
+
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.6, 3.7, 3.8]
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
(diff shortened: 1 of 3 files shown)


angerhang:
> This should replace #83 as a single clean commit. To actually achieve this, I think we need to somehow rebase master to pick out the messy series of commits in the merge from #83, and then merge this. However, this is obviously a little risky and needs to be done carefully.
> 
> @chanshing @angerhang can you confirm that you agree this is a reasonable plan?
> 
> In the future, can we not merge before checking whether it would be good to squash the commits? I was adding lots of small commits to #83 to try to fix it (because I had to run the Actions and collaborate on it), and was always planning to squash them before merging... Obviously it's not crucial from a technical point of view, but the commit history will be harder to read for future users if we let this sort of thing remain in it.

Thanks for cleaning the commit history up. I just revert the merged PR to its original state. Now it should be good to go. Apologies for not checking the commit history earlier. It's been a while since we merged a PR :D

angerhang:
The changes look fine to me.

R-Walmsley:
Thanks @angerhang, looks good to go so have merged (with a more descriptive name, and a force-push reverting one change to summariseEpoch.py within the commit which turned out to be unnecessary and probably would slow it down)

PR #87, opened 2020-11-13
Allow newer versions of sklearn

aidendoherty:
Due to a recent change in underlying dependencies, pip3 was unable to install our accelerometer package on OSX.
As a result, new users have been unable to run our software on OSX.

This change now allows the software to compile on OSX. I also tested this on Linux too.
@chanshing or @angerhang - any feedback is welcome.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,25 @@
+#
+#  Make sure we can build the java files properly
+#
+name: build
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set up JDK 1.8
+        uses: actions/setup-java@v1
+        with:
+          java-version: 1.8
+          
+      - name: Compile Parser
+        run: javac -cp java/JTransforms-3.1-with-dependencies.jar java/*.java
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,5 +1,5 @@
 #
-#  Make sure we can build the java files properly
+#  Make sure we can build the java and python files across operating systems.
 #
 name: build
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,15 +11,29 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ${{matrix.os}}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest]
 
     steps:
       - uses: actions/checkout@v2
+
+      - name: Set up Python 3.7
+        uses: actions/setup-python@v2
+        with:
+          python-version: "3.7"
+      
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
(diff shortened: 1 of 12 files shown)


angerhang:
Hey Aiden thanks for the changes.

It looks good but I would be keen for this PR to include a workflow test for the build for both MacOS and Linux so that we know (in the future), this sort of changes will work in a clean environment.

https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions has a good guide to specifying the environment being run.

The two additional tests should be similar to this https://github.com/activityMonitoring/biobankAccelerometerAnalysis/blob/master/.github/workflows/build.yml
with two extra things:
1. Install the python dependency using the command specified in the README file
2. Change of the build environments. 

Let me know if you are short on time for this, then we can create an issue and delegate it to someone else :D

angerhang (changes requested):
Good to include os-specific build tests

aidendoherty (commented):
Hi @angerhang - I completely agree with your suggestion.

I've now **attempted** to do this! Any feedback is very welcome!

.github/workflows/build.yml
```diff
@@ -11,7 +11,7 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: [ubuntu-latest, macos-latest]
```
angerhang, line 14: Think this would allow the os to be recognized. As you could see the run failed 
```
    runs-on: ${{ matrix.os }}
    strategy:
      matrix:
        os: [ubuntu-latest, macos-latest]
        python-version: [3.6, 3.7, 3.8]
```

angerhang (changes requested):

aidendoherty:
@angerhang @chanshing There is something wrong here as the build test hasn't run - so please don't accept this pull request yet. I'll have another go!

aidendoherty:
Still not working, I'll have another try after dinner!

aidendoherty:
Hi @angerhang - I have **finally** completed this (after a few failed attempts). I can now confirm that this pull request:

 1) Installs correct dependencies so that our accelerometer package now works on OSX again.  
 2) Updates the testing framework to evaluate Java and python builds on both OSX and Linux

Given all my woeful and wasted trial & error commits ... if you are happy, perhaps it would be possible to squash and merge this branch?

angerhang:
> Hi @angerhang - I have **finally** completed this (after a few failed attempts). I can now confirm that this pull request:
> 
> 1. Installs correct dependencies so that our accelerometer package now works on OSX again.
> 2. Updates the testing framework to evaluate Java and python builds on both OSX and Linux
> 
> Given all my woeful and wasted trial & error commits ... if you are happy, perhaps it would be possible to squash and merge this branch?

Everything looks great Aiden. Thx a lot for the trials and errors :D 

Ah I hate to be the devil advocate here. @chanshing introduced this which I think is a good thing so don't hate me if this drives you crazy haah :D We kinda tried to encourage a cleaner commit history when we want to merge things into the master. So the way to do this is to squash all the messy commits history into one or two? https://www.google.com/search?client=safari&rls=en&q=squash+all+commits+into+one&ie=UTF-8&oe=UTF-8 is a good guide.

aidendoherty:
Hi @angerhang - I think I've messed this up?!

If so, should I just delete this pull request and create a new one?

aidendoherty:
Cleaner pull request now submitted (#89 ).
Many thanks @angerhang for your help and advice.

PR #89, opened 2020-11-16
Fix sklearn dependency problem and new tests for python and java builds on macos and linux

aidendoherty:
After much helpful feedback from @angerhang this pull request:

1) Installs correct dependencies so that our accelerometer package now works on OSX again.
2) Updates the testing framework to evaluate Java and python builds on both OSX and Linux

======
This is an update on a previous messy pull request by me. @angerhang provided the below helpful comment in that pull request:

I would be keen for this PR to include a workflow test for the build for both MacOS and Linux so that we know (in the future), this sort of changes will work in a clean environment.

https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions has a good guide to specifying the environment being run.

The two additional tests should be similar to this https://github.com/activityMonitoring/biobankAccelerometerAnalysis/blob/master/.github/workflows/build.yml
with two extra things:

Install the python dependency using the command specified in the README file
Change of the build environments.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,6 +29,7 @@ jobs:
           python -m pip install --upgrade pip
           pip install flake8 pytest numpy scipy pandas sklearn argparse datetime statsmodels
           if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
+      
       - name: Set up JDK 1.8
         uses: actions/setup-java@v1
         with:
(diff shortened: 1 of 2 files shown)


aidendoherty:
Hi @angerhang - this should all be cleaned up now based on your helpful feedback at #87 . When you get a chance, can you review this pull request?

angerhang (approved):
@aidendoherty lovely. Looks good to me

Issue #90, opened 2020-11-24
get_clips and get_total_reads produce JSON non-seriarizable 


chanshing:
@smallsr Scott Small encountered errors while processing his files, which I traced back to `get_clips` and `get_total_reads` https://github.com/activityMonitoring/biobankAccelerometerAnalysis/commit/42871e33f5897aeadba5fcfe8bb79d25d5e5e3ef

The problem is that `np.sum(e['clipsBeforeCalibr'])` etc. sometimes gives us a `numpy.float64` type (when there are NaNs in the column, e.g. missing values, interrupts), and sometimes a `numpy.int64` type (when no NaNs). Turns out the `json` package can't serialize `numpy.int64` types for some reason, resulting in `TypeError: Object of type int64 is not JSON serializable` error.

A quick fix is as follows

```python
def get_clips(e, epochPeriod, summary):
    summary['clipsBeforeCalibration'] = np.sum(e['clipsBeforeCalibr']).item()
    summary['clipsAfterCalibration'] = np.sum(e['clipsAfterCalibr']).item()

def get_total_reads(e, epochPeriod, summary):
    summary['totalReads'] = np.sum(e['rawSamples']).item()
```
Basically, adding a `.item()` call to convert from `numpy.*` type to native type (e.g. `numpy.int64` to native `int`, and `numpy.float64` to native `float`). Alternatively one can use `accUtils.formatNum` as it is done in `summariseEpoch.writeMovementSummaries()`.

I can submit a PR later unless someone is free to do it right away 👯

PR #91, opened 2020-11-24
fix bug in get_clips and get_total_reads

chanshing:
This is to fix https://github.com/activityMonitoring/biobankAccelerometerAnalysis/issues/90
@smallsr encountered the error `TypeError: Object of type int64 is not JSON serializable` when processing his files, happening when the program tries to write out the summary in JSON format. 
For some reason, `json` package can't serialize `np.int64`, but `np.float64` is OK 😕 
The sums in `get_clips` and `get_total_reads` produce `np.float64` type when there are NaNs in the columns (happens when there are interrupts or nonwear), but otherwise are of type `np.int64` and that's when it fails. Our sample cwa file has nonwear and interrupts so the bug didn't manifest. 
A quick fix is to convert to native types using `.item()`

### DIFF
--- a/accelerometer/summariseEpoch.py
+++ b/accelerometer/summariseEpoch.py
@@ -157,11 +157,11 @@ def getActivitySummary(epochFile, nonWearFile, summary,
     return e, labels
 
 def get_clips(e, epochPeriod, summary):
-    summary['clipsBeforeCalibration'] = np.sum(e['clipsBeforeCalibr'])
-    summary['clipsAfterCalibration'] = np.sum(e['clipsAfterCalibr'])
-    
+    summary['clipsBeforeCalibration'] = e['clipsBeforeCalibr'].sum().item()
+    summary['clipsAfterCalibration'] = e['clipsAfterCalibr'].sum().item()
+
 def get_total_reads(e, epochPeriod, summary):
-    summary['totalReads'] = np.sum(e['rawSamples'])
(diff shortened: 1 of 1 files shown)


chanshing:
Sorry, I now realize the bugfix branch was created on the main repo instead of on my fork. I was using the Github interface to do it quickly and didn't notice it. I hope this is OK otherwise.

angerhang (changes requested):
Looks good but is it possible to provide a test via Actions to compare the generated summary file with a reference one? It could be part of one of an existing workflows so should be easy to implement.

chanshing:
> Looks good but is it possible to provide a test via Actions to compare the generated summary file with a reference one? It could be part of one of an existing workflows so should be easy to implement.

Can do. I will try to have it by end of the week since I haven't got my hands on this whole Action thing yet.

R-Walmsley (approved):
Missed seeing this before, but this fix looks great!

chanshing:
> Missed seeing this before, but this fix looks great!

Great :) @angerhang @R-Walmsley I still need to add the test but feel free to do that if you want.

R-Walmsley:
@angerhang, how would you feel about merging this and then adding the test separately? I would like to use this soon...

angerhang:
> @angerhang, how would you feel about merging this and then adding the test separately? I would like to use this soon...

If that's the case, sure but let's create an issue to remind us to add the tests :D

angerhang (approved):

Issue #93, opened 2020-11-26
AccPlot broken for locator mismatch 


angerhang:
Likely due to a recent update in matplotlib, the accplot.py fails to plot with the following error msg:

```
ValueError: The number of FixedLocator locations (7), usually from a call to set_ticks, does not match the number of ticklabels (6)
```

aidendoherty:
Fixed by @chanshing in #104

Issue #92, opened 2020-11-26
Specify python package versions in testing


angerhang:
Currently, our installation files do not specify python package which from time to time miss the issues raised by a package update. Should be an easy fix for the installation procedure. Maybe it is time to use conda to take care of that.

R-Walmsley:
Addressed in PR #109

Issue #20, opened 2020-03-20
Can't import accelerometer No module named 'numpy.testing.decorators'


muschellij2:
Installed via the REAMDE:
```
MacBook-Pro-9:biobankAccelerometerAnalysis johnmuschelli$ python3
Python 3.5.1 (v3.5.1:37a07cee5969, Dec  5 2015, 21:12:44) 
[GCC 4.2.1 (Apple Inc. build 5666) (dot 3)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import accelerometer
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/johnmuschelli/Dropbox/Packages/biobankAccelerometerAnalysis/accelerometer/__init__.py", line 1, in <module>
    import accelerometer.accClassification
  File "/Users/johnmuschelli/Dropbox/Packages/biobankAccelerometerAnalysis/accelerometer/accClassification.py", line 6, in <module>
    import pandas as pd
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/pandas/__init__.py", line 56, in <module>
    import pandas.util.testing
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/pandas/util/testing.py", line 22, in <module>
    from numpy.testing.decorators import slow     # noqa
ImportError: No module named 'numpy.testing.decorators'
```
May be related to:  https://github.com/apache/incubator-mxnet/issues/17154

chanshing:
It seems to be related to this:
https://stackoverflow.com/questions/59474533/modulenotfounderror-no-module-named-numpy-testing-nosetester
Please can you show your `scipy` and `numpy` versions? Try upgrading `scipy`.

muschellij2:
Still failing after updating `scipy`:
```
MacBook-Pro-9:biobankAccelerometerAnalysis johnmuschelli$ pip freeze | grep numpy
numpy==1.18.1
MacBook-Pro-9:biobankAccelerometerAnalysis johnmuschelli$ pip freeze | grep scipy
scipy==1.1.0
```

chanshing:
Can you try downgrading `numpy` to `1.18`

muschellij2:
Downgraded, same thing.

chanshing:
Can you try upgrading `pandas`? Or create a fresh python environment and install `pandas` first.

[...]

PR #94, opened 2020-11-26
fix plotting and add tests

angerhang:
Temporary fix for https://github.com/activityMonitoring/biobankAccelerometerAnalysis/issues/93.

Added a test for plotting.

### DIFF
--- a/accPlot.py
+++ b/accPlot.py
@@ -195,7 +195,7 @@ def plotTimeSeries(
             datetime.combine(day + timedelta(days=1), time(0, 0, 0, 0))))
         ax.set_xticks(pd.date_range(start=datetime.combine(day,time(0, 0, 0, 0)),
             end=datetime.combine(day + timedelta(days=1), time(0, 0, 0, 0)),
-            freq='4H'))
+            freq='4H')[:-1])
         ax.set_xticks(pd.date_range(start=datetime.combine(day,time(0, 0, 0, 0)),
             end=datetime.combine(day + timedelta(days=1), time(0, 0, 0, 0)),
             freq='1H'), minor=True)
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -27,7 +27,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install flake8 pytest numpy scipy pandas sklearn argparse datetime statsmodels
+          pip install .
           if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
       
       - name: Set up JDK 1.8
--- a/.github/workflows/regression_test_summary.yml
+++ b/.github/workflows/regression_test_summary.yml
@@ -23,7 +23,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install flake8 pytest numpy scipy pandas sklearn argparse datetime statsmodels
+        pip install .
         if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
    # - name: Lint with flake8
    #   run: |
@@ -43,6 +43,5 @@ jobs:
       with:
          original: referenceFiles/epoch.csv.gz
          mirror: data/sample-epoch.csv.gz
-
-
-
+    - name: Plot output
+      run: python accPlot.py data/sample-timeSeries.csv.gz data/sample-plot.png


chanshing:
Hi Hang, on a second thought wouldn't it be better to just enforce people to use matplotlib3.2.2 (or whatever latest that works)?

angerhang:
> Hi Hang, on a second thought wouldn't it be better to just enforce people to use matplotlib3.2.2 (or whatever latest that works)?

This is tricky because what this would mean is that we would need to hand pick version number for dependencies which is more work. I have a second look, the reason why CI fails was because `matplotlib` wasn't installed at all in our build tests. Now it is fixed and should always test against the latest release.

angerhang:
Should be good to go know that I've just leaned the commit history :D

angerhang:
@chanshing  @R-Walmsley @aidendoherty  can we have a quick a look at this outstanding PR :D

chanshing:
Hi Hang can you explain what this does to fix the issue? Also how to ensure this doesn't happen every time matplotlib, pandas or other packages get updated?

R-Walmsley:
> > Hi Hang, on a second thought wouldn't it be better to just enforce people to use matplotlib3.2.2 (or whatever latest that works)?
> 
> This is tricky because what this would mean is that we would need to hand pick version number for dependencies which is more work. I have a second look, the reason why CI fails was because `matplotlib` wasn't installed at all in our build tests. Now it is fixed and should always test against the latest release.

I don't understand why we wouldn't force version number? Isn't there a way to force >3.2.2 (or as appropriate). So if someone doesn't have it installed they install most recent version, but if someone has an old version installed it will flag that?

angerhang:
> > > Hi Hang, on a second thought wouldn't it be better to just enforce people to use matplotlib3.2.2 (or whatever latest that works)?
> > 
> > 
> > This is tricky because what this would mean is that we would need to hand pick version number for dependencies which is more work. I have a second look, the reason why CI fails was because `matplotlib` wasn't installed at all in our build tests. Now it is fixed and should always test against the latest release.
> 
> I don't understand why we wouldn't force version number? Isn't there a way to force >3.2.2 (or as appropriate). So if someone doesn't have it installed they install most recent version, but if someone has an old version installed it will flag that?

Well, both of you are raising valid concerns here. After some googling around, there doesn't seem to be a universally accepted standard of testing dependencies together with CI. Having said that, I think we can make certain that our `requirement.txt` list has version numbers. And probably we should use that for CI. But that should be a separate issue but how to enforce version number for the dependencies we have.

The changes made in this PR is letting us install using the `setup.py` for installations. Previously, it was simply installing a few hand-select packages.

accPlot.py
```diff
@@ -195,7 +195,7 @@ def plotTimeSeries(
             datetime.combine(day + timedelta(days=1), time(0, 0, 0, 0))))
         ax.set_xticks(pd.date_range(start=datetime.combine(day,time(0, 0, 0, 0)),
             end=datetime.combine(day + timedelta(days=1), time(0, 0, 0, 0)),
-            freq='4H'))
+            freq='4H')[:-1])
```
chanshing, line 198: The problem I have with this quick fix is that I'm not sure we want to "fix" the issue because it seems to be a problem with matplotlib rather than our code. The same issue was posted here: https://stackoverflow.com/questions/62953704/valueerror-the-number-of-fixedlocator-locations-5-usually-from-a-call-to-set Top comment:

> Let me guess. You used matplotlib version 3.2.2 and today created a new environment with matplotlib 3.3.0. Something is broken in new version although I didn't find an exact bug report yet. I suggest to revert back to 3.2.2 until this is resolved.

I agree with this solution. The code worked for _all_ previous matplotlib versions and stopped working only for the latest 3.3.* 

It's not clear why the bug happens and whether your quick fix will break for people using previous matploblib versions.
I suggest `conda install matplotlib=3.2.2` and be clear about package dependencies of our software (I think this is standard in other projects). As always, feel free to disagree!

angerhang: Thanks for the suggestion. It also works with 3.2.2 :D 

But again, fixing the package dependencies is not what this PR is for. That is a separate issue https://github.com/activityMonitoring/biobankAccelerometerAnalysis/issues/92 and is still outstanding : (

chanshing: Yes, it works for 3.2.2 fortunately, but I haven't tried for even older versions. I think we should just wait until matplotlib fixes this bug.

angerhang: It is going to be a bit tricky because this is this DREAMS project that will need to have the plotting for the trial lolol. Maybe I will just ask them to use my fork instead. There will be too many changes and new things coming and I don't think we will be able to finish all of that in time for PR. Whilst this is not ideal, that's the best that we can do right now.

chanshing: Yes I agree. If your collaborators are adamant of using matplotlib=3.3.* we can point them to your branch.

angerhang: On a related note, I suspect this is more of a change of behaviour than a bug TBH. At least, till today, we are not seeing any bug report yet on Matplotlib repo. Previously, the fixlocator can have different lengths from the axises, now it cannot within 3.3.3. It is probably the `right` behaviour from my current understanding.

Yeah the issue is that the research assistants will be less technically sophisticated and are not well-versed in dependency management.

angerhang: Sure let's discuss this during the next meeting and hopefully when the new dependency management tool is being resolved. This PR should settle properly. The longer it waits, the more time it would take to revisit this PR which is not ideal : (

chanshing: > Yeah the issue is that the research assistants will be less technically sophisticated and are not well-versed in dependency management.

Agree, that's why we need to improve our installation process to make sure we are grabbing the right versions. Today is matplotlib but tomorrow might be pandas. Last time was sklearn and joblib. I will look into this too so we can see what's the best way to approach it. Otherwise we shall expect more of these https://github.com/activityMonitoring/biobankAccelerometerAnalysis/issues/20 in the future.

chanshing:
Hi @angerhang 

I decided to dig deeper into this. So as you said in `matplotlib=3.3+` they now enforce you to provide the _same_ number _labels_ as there are _ticks_, whereas in `matplotlib=3.2-` you could be more flexible. Let's first look at the traceback:

```bash
Traceback (most recent call last):
  File "accPlot.py", line 262, in <module>
    main()  # Standard boilerplate to call the main() function to begin the program.
  File "accPlot.py", line 73, in main
    imputedLabelsHeight=args.imputedLabelsHeight)
  File "accPlot.py", line 246, in plotTimeSeries
    ax_list[0].set_xticklabels(hrLabels)
  File "/path/to/anaconda3/lib/python3.7/site-packages/matplotlib/axes/_base.py", line 63, in wrapper
    return get_method(self)(*args, **kwargs)
  File "/path/to/anaconda3/lib/python3.7/site-packages/matplotlib/cbook/deprecation.py", line 451, in wrapper
    return func(*args, **kwargs)
  File "/path/to/anaconda3/lib/python3.7/site-packages/matplotlib/axis.py", line 1793, in _set_ticklabels
    return self.set_ticklabels(labels, minor=minor, **kwargs)
  File "/path/to/anaconda3/lib/python3.7/site-packages/matplotlib/axis.py", line 1715, in set_ticklabels
    "The number of FixedLocator locations"
ValueError: The number of FixedLocator locations (7), usually from a call to set_ticks, does not match the number of ticklabels (6).
```

So the offending line is `ax_list[0].set_xticklabels(hrLabels)` where we try to set the labels 0am, 4am, 8am, 12am, 4pm, 8pm (so 6 labels). But turns out we have 7 ticks (0, 4, 8, 12, 4, 8, and 0 again), but so far we never labeled the last 0:

Original output:
image

As you see the last tick (midnight) is not labeled. In `matplotlib=3.2-` it didn't complain about the missing label, but in 3.3+ it now does. So I agree with you that this is not a bug (although they are breaking people's code!) but rather a change in behavior that is likely to stay permanent I think.

Now I propose that we fix `hrLabels` to include the last midnight label, instead of your proposed solution to hack the number of ticks.

Here's a plot from your solution:

image

You might need magnifying glasses to spot the difference (and sorry for being extremely picky right now 🤓 ). The difference is in the length of the last tick (below where the last 0 should be). But this is not at all the reason I don't like this solution. 

The reason is that the fix looks asymmetric and it's not immediately clear why it is done:

```python
        # Lines 195-200 in accPlot.py
        ax.set_xticks(pd.date_range(start=datetime.combine(day,time(0, 0, 0, 0)),
            end=datetime.combine(day + timedelta(days=1), time(0, 0, 0, 0)),
            freq='4H')[:-1])  # <-- WHY? --my future self 5 years from now
        ax.set_xticks(pd.date_range(start=datetime.combine(day,time(0, 0, 0, 0)),
            end=datetime.combine(day + timedelta(days=1), time(0, 0, 0, 0)),
            freq='1H'), minor=True)  # <-- and here?
```

So I suggest we simply edit `hrLabels` instead, which actually addresses the issue more directly:

```python
    # Before
    hours2Display = range(0, 24, 4)
    hrLabels = [(str(hr) + 'am') if hr<=12 else (str(hr-12) + 'pm') for hr in hours2Display]  # ['0am', '4am', '8am', '12am', '4pm', '8pm']

    # Proposed
    hrLabels = ['0am', '4am', '8am', '12am', '4pm', '8pm', '12pm']
```

Output:
image


### Regarding 12am vs 12pm
I'm always confused by 12am/12pm and luckily it's not just me: https://en.wikipedia.org/wiki/12-hour_clock#Confusion_at_noon_and_midnight
So since we are at it, may I also suggest the alternative:

```python
    # Proposed
    hrLabels = ['midnight', '4am', '8am', 'noon', '4pm', '8pm', 'midnight']
```
Output
image

Open to suggestions @angerhang @aidendoherty @R-Walmsley

R-Walmsley:
This looks great @chanshing, and your reasoning is very clear to me at least. 
The only part I slightly disagree on is the midnight thing. While I also always get confused by this, I kind of like the fact it's numeric and also at least somewhat user-language-agnostic. 
Another alternative would be using the 24 hour clock (ie. 00:00, 04:00, 08:00, 12:00, 16:00, 20:00, 00:00)? Rationale: it's most general, unambiguous. Also widely used (in many places more than the 12-12 clock).

chanshing:
> This looks great @chanshing, and your reasoning is very clear to me at least.
> The only part I slightly disagree on is the midnight thing. While I also always get confused by this, I kind of like the fact it's numeric and also at least somewhat user-language-agnostic.
> Another alternative would be using the 24 hour clock (ie. 00:00, 04:00, 08:00, 12:00, 16:00, 20:00, 00:00)? Rationale: it's most general, unambiguous. Also widely used (in many places more than the 12-12 clock).

+1 for the 24 hour clock! Thanks for the feedback 👾

aidendoherty:
Team has agreed to close this.

PR #95, opened 2020-12-13
avoid error message when epoch folder manually changed 

R-Walmsley:
When `--processInputFile` is `False`, the epochFolder may be manually changed to a folder with previously processed data. write permissions are not required on this folder (and it is likely they are not available, as this data should not be overwritten). 
This PR should allow this folder to have different permissions in the `--processInputFile False` case.

### DIFF
--- a/accProcess.py
+++ b/accProcess.py
@@ -288,12 +288,17 @@ def main():
 
     # Check if we can write to the output folders
     for path in [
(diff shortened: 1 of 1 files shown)


angerhang (approved):
Looks good to me!

chanshing:
Looks good to me. Feel free to merge 👯

chanshing (approved):

PR #99, opened 2020-12-29
fix bug in interrupts; deprecation; refactor

chanshing:
This is to fix a bug in `get_interrupts` that results in NaN in `summary['errs-interrupt-mins']`. The problem is that pandas `.diff()` creates a NaN in the first row -- e.g. see these examples https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.diff.html
This makes `summary['errs-interrupt-mins']` always NaN. Apparently no one noticed it so far.
Second, I also removed `keep_tz` flag which is deprecated https://pandas.pydata.org/pandas-docs/stable/whatsnew/v1.0.0.html#deprecations
Finally, refactor for consistent spacing between functions.

### DIFF
--- a/accelerometer/summariseEpoch.py
+++ b/accelerometer/summariseEpoch.py
@@ -156,9 +156,13 @@ def getActivitySummary(epochFile, nonWearFile, summary,
     # Return physical activity summary
     return e, labels
 
+
+
 def get_clips(e, epochPeriod, summary):
     summary['clipsBeforeCalibration'] = e['clipsBeforeCalibr'].sum().item()
     summary['clipsAfterCalibration'] = e['clipsAfterCalibr'].sum().item()
+
+
 
 def get_total_reads(e, epochPeriod, summary):
     summary['totalReads'] = e['rawSamples'].sum().item()
--- a/accelerometer/summariseEpoch.py
+++ b/accelerometer/summariseEpoch.py
@@ -163,6 +163,8 @@ def get_clips(e, epochPeriod, summary):
 def get_total_reads(e, epochPeriod, summary):
     summary['totalReads'] = e['rawSamples'].sum().item()
 
+
+
 def get_interrupts(e, epochPeriod, summary):
     """Identify if there are interrupts in the data recording
 
@@ -175,12 +177,9 @@ def get_interrupts(e, epochPeriod, summary):
     """
 
     epochNs = epochPeriod * np.timedelta64(1, 's')
-    interrupts = np.where(e.index.to_series(keep_tz=True).diff() > epochNs)[0]
+    interrupts = np.where(e.index.to_series().diff() > epochNs)[0]
     # Get duration of each interrupt in minutes
-    interruptMins = []
-    for i in interrupts:
-        interruptMins.append(e.index[i-1:i+1].to_series(keep_tz=True).diff() /
-         np.timedelta64(1, 'm'))
+    interruptMins = [e.index[i-1:i+1].to_series().diff().sum(skipna=True).seconds / 60 for i in interrupts]
     # Record to output summary
     summary['errs-interrupts-num'] = len(interruptMins)
     summary['errs-interrupt-mins'] = accUtils.formatNum(np.sum(interruptMins), 1)


angerhang:
I think you are having the same issues as I am because the current installation script doesn't really install everything. I think some changes introduced here could fix this: https://github.com/activityMonitoring/biobankAccelerometerAnalysis/pull/94/files

chanshing:
> I think you are having the same issues as I am because the current installation script doesn't really install everything. I think some changes introduced here could fix this: https://github.com/activityMonitoring/biobankAccelerometerAnalysis/pull/94/files

Hi Hang, Merry Christmas! 
sklearn 0.24 is finally here and it's breaking our code https://scikit-learn.org/stable/whats_new/v0.24.html 
To be fair, it has been warning us for some time already:

```bash
/path/to/username/anaconda3/lib/python3.7/site-packages/sklearn/utils/deprecation.py:143: FutureWarning: The sklearn.ensemble.forest module is  deprecated in version 0.22 and will be removed in version 0.24. The corresponding classes / functions should instead be imported from sklearn.ensemble. Anything that cannot be imported from sklearn.ensemble is now part of the private API.
  warnings.warn(message, FutureWarning)
```

I had fixed this somewhere sometime before, basically they refactored `forest` to `_forest` to make it "private", so we need to import `_forest` now instead.

chanshing:
> > I think you are having the same issues as I am because the current installation script doesn't really install everything. I think some changes introduced here could fix this: https://github.com/activityMonitoring/biobankAccelerometerAnalysis/pull/94/files
> 
> Hi Hang, Merry Christmas!
> sklearn 0.24 is finally here and it's breaking our code https://scikit-learn.org/stable/whats_new/v0.24.html
> To be fair, it has been warning us for some time already:
> 
> ```shell
> /path/to/username/anaconda3/lib/python3.7/site-packages/sklearn/utils/deprecation.py:143: FutureWarning: The sklearn.ensemble.forest module is  deprecated in version 0.22 and will be removed in version 0.24. The corresponding classes / functions should instead be imported from sklearn.ensemble. Anything that cannot be imported from sklearn.ensemble is now part of the private API.
>   warnings.warn(message, FutureWarning)
> ```
> 
> I had fixed this somewhere sometime before (some other code using sklearn RF). Basically they refactored `forest` to `_forest` to make it "private", so we need to import `_forest` now instead.
> 
> Package dependency strikes once again!

This is the error from Github Actions:

```bash
2020-12-29 22:39:06 (2.25 MB/s) - ‘activityModels/labelled-acc-epochs.csv’ saved [16141812/16141812]

Traceback (most recent call last):
  File "accProcess.py", line 3, in <module>
    import accelerometer.accUtils
  File "/home/runner/work/biobankAccelerometerAnalysis/biobankAccelerometerAnalysis/accelerometer/__init__.py", line 1, in <module>
    import accelerometer.accClassification
  File "/home/runner/work/biobankAccelerometerAnalysis/biobankAccelerometerAnalysis/accelerometer/accClassification.py", line 9, in <module>
    import sklearn.ensemble.forest as forest
ModuleNotFoundError: No module named 'sklearn.ensemble.forest'
Error: Process completed with exit code 1.
```

This is so nice. CI paying off :+1:

angerhang:
> > I think you are having the same issues as I am because the current installation script doesn't really install everything. I think some changes introduced here could fix this: https://github.com/activityMonitoring/biobankAccelerometerAnalysis/pull/94/files
> 
> Hi Hang, Merry Christmas!
> sklearn 0.24 is finally here and it's breaking our code https://scikit-learn.org/stable/whats_new/v0.24.html
> To be fair, it has been warning us for some time already:
> 
> ```shell
> /path/to/username/anaconda3/lib/python3.7/site-packages/sklearn/utils/deprecation.py:143: FutureWarning: The sklearn.ensemble.forest module is  deprecated in version 0.22 and will be removed in version 0.24. The corresponding classes / functions should instead be imported from sklearn.ensemble. Anything that cannot be imported from sklearn.ensemble is now part of the private API.
>   warnings.warn(message, FutureWarning)
> ```
> 
> I had fixed this somewhere sometime before (some other code using sklearn RF). Basically they refactored `forest` to `_forest` to make it "private", so we need to import `_forest` now instead.
> 
> Package dependency strikes once again! 😂

Merry Christmas Shing. There is nothing to do here in the UK right now but hopefully, your home is better.
Yeah, we need to resolve this issue during our next meeting but for the time being, should we just do what's recommended by importing `_forest`? It is breaking my fork too lolol.

chanshing:
> > > I think you are having the same issues as I am because the current installation script doesn't really install everything. I think some changes introduced here could fix this: https://github.com/activityMonitoring/biobankAccelerometerAnalysis/pull/94/files
> > 
> > 
> > Hi Hang, Merry Christmas!
> > sklearn 0.24 is finally here and it's breaking our code https://scikit-learn.org/stable/whats_new/v0.24.html
> > To be fair, it has been warning us for some time already:
> > ```shell
> > /path/to/username/anaconda3/lib/python3.7/site-packages/sklearn/utils/deprecation.py:143: FutureWarning: The sklearn.ensemble.forest module is  deprecated in version 0.22 and will be removed in version 0.24. The corresponding classes / functions should instead be imported from sklearn.ensemble. Anything that cannot be imported from sklearn.ensemble is now part of the private API.
> >   warnings.warn(message, FutureWarning)
> > ```
> > 
> > 
> > I had fixed this somewhere sometime before (some other code using sklearn RF). Basically they refactored `forest` to `_forest` to make it "private", so we need to import `_forest` now instead.
> > Package dependency strikes once again! joy
> 
> Merry Christmas Shing. There is nothing to do here in the UK right now but hopefully, your home is better.
> Yeah, we need to resolve this issue during our next meeting but for the time being, should we just do what's recommended by importing `_forest`? It is breaking my fork too lolol.

Yup, I guess that's the solution. I will look into how to specify the package versions to use in Actions.

angerhang:
> https://docs.github.com/en/free-pro-team@latest/actions/guides/building-and-testing-python

That's great. In fact, we already have github actions to test different python versions. We probably shouldn't directly handle dependency manage using Github actions but that specify the package version via `requirement.txt` or `setup.py` whatever we use to install in general. Again, the current build in Github actions needs to be updated accordingly as I suggested previously :D

chanshing:
> > https://docs.github.com/en/free-pro-team@latest/actions/guides/building-and-testing-python
> 
> That's great. In fact, we already have github actions to test different python versions. We probably shouldn't directly handle dependency manage using Github actions but that specify the package version via `requirement.txt` or `setup.py` whatever we use to install in general. Again, the current build in Github actions needs to be updated accordingly as I suggested previously :D



> > https://docs.github.com/en/free-pro-team@latest/actions/guides/building-and-testing-python
> 
> That's great. In fact, we already have github actions to test different python versions. We probably shouldn't directly handle dependency manage using Github actions but that specify the package version via `requirement.txt` or `setup.py` whatever we use to install in general. Again, the current build in Github actions needs to be updated accordingly as I suggested previously :D

Is this the approach you are suggestion (from the link)?:

```bash
name: Python package

on: [push]

jobs:
  build:

    runs-on: ubuntu-latest
    strategy:
      matrix:
        python-version: [2.7, 3.5, 3.6, 3.7, 3.8]

    steps:
    - uses: actions/checkout@v2
    - name: Set up Python ${{ matrix.python-version }}
      uses: actions/setup-python@v2
      with:
        python-version: ${{ matrix.python-version }}
    - name: Install dependencies
      run: |
        python -m pip install --upgrade pip
        pip install flake8 pytest
        if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
    - name: Lint with flake8
      run: |
        # stop the build if there are Python syntax errors or undefined names
        flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
        # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
        flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
    - name: Test with pytest
      run: |
        pytest
```

Or are you suggesting something else?

angerhang:
> ```shell
>         pip install flake8 pytest
> ```

I was suggesting something else. This script contains a lot more checks than what we could deal with right now. It would need a lot of refactoring until we can reach that point.

I was referring to changing https://github.com/activityMonitoring/biobankAccelerometerAnalysis/blob/61b36d1ae6212de93ec10a3cb164d40b2f6b5ace/.github/workflows/build.yml#L30

into 
```
pip install .
```
so that we can directly specify the package version in either `setup.py` or `requirement.txt`. Our current setup manually installs some of the dependencies and doesn't use `setup.py` or `requirement.txt` to install in Github actions. Hope that makes sense :D

chanshing:
> > ```shell
> >         pip install flake8 pytest
> > ```
> 
> I was suggesting something else. This script contains a lot more checks than what we could deal with right now. It would need a lot of refactoring until we can reach that point.
> 
> I was referring to changing
> 
> https://github.com/activityMonitoring/biobankAccelerometerAnalysis/blob/61b36d1ae6212de93ec10a3cb164d40b2f6b5ace/.github/workflows/build.yml#L30
> 
> into
> 
> ```
> pip install .
> ```
> 
> so that we can directly specify the package version in either `setup.py` or `requirement.txt`. Our current setup manually installs some of the dependencies and doesn't use `setup.py` or `requirement.txt` to install in Github actions. Hope that makes sense :D

Can you explain what `pip install .` does? (sorry I am more used to conda). Does it automatically look for `requirements.txt` and install what's in there? I thought this line from the example is doing that explicitly:

```bash
        pip install flake8 pytest
        if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
```

Is that correct?

angerhang:
> ```shell
> ```shell
>         pip install flake8 pytest
> ```
> 
> 
> 
> I was suggesting something else. This script contains a lot more checks than what we could deal with right now. It would need a lot of refactoring until we can reach that point.
> I was referring to changing
> https://github.com/activityMonitoring/biobankAccelerometerAnalysis/blob/61b36d1ae6212de93ec10a3cb164d40b2f6b5ace/.github/workflows/build.yml#L30
> 
> into
> pip install .
> 
> so that we can directly specify the package version in either setup.py or requirement.txt. Our current setup manually installs some of the dependencies and doesn't use setup.py or requirement.txt to install in Github actions. Hope that makes sense :D
> 
> Can you explain what pip install . does? (sorry I am more used to conda). Does it automatically look for requirements.txt and install what's in there? I thought this line from the example is doing that explicitly:
> ```shell
>         pip install flake8 pytest
>         if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
> ```
> 
> 
> Is that correct?
> Edit: Ok so https://stackoverflow.com/questions/39023758/what-does-pip-install-dot-mean answers my question.
> Is it better to use the example from Github or is pip install . better?
> ```

Yeah it has to do with this outstanding issue https://github.com/activityMonitoring/biobankAccelerometerAnalysis/issues/92.
In practice, it probably doesn't matter how we do it but we just need to choose one. Currently, we only have `setup.py` but no `requirement.txt`. If we opt for conda, that could work too :D

R-Walmsley:
Hi @angerhang @chanshing ... Could I just check I understand the decision here? Is it to: 
1) temporarily fix it by forcing particular versions of the dependencies in the testing 
2) also long-term fix by changing it to work correctly with the newer versions of `sklearn`? 

If so, could we add something to the documentation to make clear that you need to be using the package versions as specified in `requirements.txt` (if that is the case)?

PR #104, opened 2021-01-14
fix hrLabels missing last tick; also use 24 format

chanshing:
Follow-up of https://github.com/activityMonitoring/biobankAccelerometerAnalysis/pull/94
This fixes the missing last tick label in `hrLabels` in accPlot.py, which was breaking with the latest matplotlib versions. 
Also uses the 24-hour format rather than AM/PM.

New output:

sample


aidendoherty:
Great work @chanshing

aidendoherty (approved):
Looks good

Issue #103, opened 2021-01-12
5 second of epoch leads to all timeseries with imputed=1


xiangnandang:
https://github.com/activityMonitoring/biobankAccelerometerAnalysis/blob/61b36d1ae6212de93ec10a3cb164d40b2f6b5ace/accelerometer/accUtils.py#L476

if 5 second epoch is used, `avgArmAngelAbsDiff` from epoch file / df will always be `na`, thus `imputed` in timeseries file will always be `1`

This seemed to work before: `e['imputed'] = np.isnan(e['acc']).astype(int)`, but there might be reason I'm not aware to use current `e_new['imputed'] = e.isna().any(1).astype('int')`

angerhang:
Great question @xiangnandang. Thanks for reporting.

@chanshing this behavior is introduced in https://github.com/activityMonitoring/biobankAccelerometerAnalysis/commit/379ebb2025f2ed44334c631288b85b2f3e660e6b. For the imputed field, shouldn't the behavior be that if only `acc` is NA then, we should mark an epoch as imputed instead of the whole frame? Is this because if there is any features of an epoch that cannot be computed, we would have to impute that epoch?

chanshing:
> Great question @xiangnandang. Thanks for reporting.
> 
> @chanshing this behavior is introduced in [379ebb2](https://github.com/activityMonitoring/biobankAccelerometerAnalysis/commit/379ebb2025f2ed44334c631288b85b2f3e660e6b). For the imputed field, shouldn't the behavior be that if only `acc` is NA then, we should mark an epoch as imputed instead of the whole frame? Is this because if there is any features of an epoch that cannot be computed, we would have to impute that epoch?

Hmmm AFAIK we don't impute features other than the mean acceleration. But we do sort of impute the predicted activities -- we average the predictions for the other days for that same epoch. So if any feature was NaN, then the prediction would be NaN too for that epoch, then it would need to be imputed.

If I understand correctly, `avgArmAngelAbsDiff` is a new feature that is not used for activity classification, so `avgArmAngelAbsDiff` may be NaN but the classifier is still working. So `e_new['imputed']` should actually look at the features used by the classifier rather than the whole.

xiangnandang:
Thank you both for responding so quickly!
@chanshing , yes, your understanding is correct!

chanshing:
> Thank you both for responding so quickly!
> @chanshing , yes, your understanding is correct!

[...]

PR #105, opened 2021-01-15
fix 'imputed' column in timeSeries file

chanshing:
To fix https://github.com/activityMonitoring/biobankAccelerometerAnalysis/issues/103

The `imputed` column in the output timeSeries.csv file should be based on `acc` and `labels` rather than the whole row.

@angerhang I think this issue was assigned to you, but since the fix is on code I wrote I just went ahead and did it. 
There's still an issue with `avgArmAngelAbsDiff` which returns NaN for exactly `--epochPeriod 5`. I'll leave that fix to you.

### DIFF
--- a/accelerometer/accUtils.py
+++ b/accelerometer/accUtils.py
@@ -473,7 +473,7 @@ def writeTimeSeries(e, labels, tsFile):
 
     e_new = pd.DataFrame(index=e.index)
     e_new.index.name = 'time'
-    e_new['imputed'] = e.isna().any(1).astype('int')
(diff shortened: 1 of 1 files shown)


angerhang:
Looks good to me.

Since this branch hasn't pulled the latest changes to fix the sklearn, CI fails but should be safe to accept.

angerhang (approved):

chanshing:
Just rebased. Checks passing now.
Merging this as @angerhang approved and he was the one closely following https://github.com/activityMonitoring/biobankAccelerometerAnalysis/issues/103

PR #108, opened 2021-01-19
Changes to use scikit-learn 0.24.0

R-Walmsley:
I think these is the minimal set of changes needed to address Issue #101. 

- imports _forest as private function 
- gives _parallel_build_trees_ an (unused) n_bootstrap_samples argument to match what it is passed from other functions 
- _parallel_build_trees_ does not change at all internally from previous versions of biobankAccelerometerAnalysis. As far as I can tell, this seems compatible with functions in 0.24.0 which call it - but may be risky when working with, and redefining, private functions. 

Feedback very welcome!

Issue #101, opened 2021-01-01
Sklearn breaks classification model 


angerhang:
As mentioned in https://github.com/activityMonitoring/biobankAccelerometerAnalysis/pull/99, sklearn 0.24 breaks forest. When one imports the ``_forest`` package, the current downloaded model is not compatible with the newly imported model for classification. 

Ideally, we perhaps should retrain the model with the right class and use sklearn as the latest supported version.

R-Walmsley:
Addressed in PR #108.

PR #109, opened 2021-01-19
Implement installation with requirements.txt

R-Walmsley:
This PR implements installation using `requirements.txt`, addressing discussions under Issues #101 and #92

If this were merged, changes would be: 
- Default installation instructions use requirements.txt, rather than setup.py 
- Version numbers much more rigid than previously (currently entirely rigid, but can be relaxed if requested). Current version numbers are based on a `pip freeze` of a known working version.
- Adds a `pip upgrade` step to avoid issues due to outdated pip version

Notes: 
- setup.py is not removed yet 
- workflows are not yet updated to use only `requirements.txt`
- to run correctly, this would need to be merged after #108 as it installs scitkit-learn 0.24.0

This represents one possible solution to specifying package versions - happy to hear comments, including if you don't like the look of it at all!


angerhang (approved):

chanshing:
Hi Rosemary, I thought we were sticking to sklearn 0.23 for now since 0.24 is not yet available in conda? Anaconda is the most popular manager for data scientists and I'm afraid it might break for many users.

angerhang:
> Hi Rosemary, I thought we were sticking to sklearn 0.23 for now since 0.24 is not yet available in conda? Anaconda is the most popular manager for data scientists and I'm afraid it might break for many users.

I think this is probably fine because I normally use Conda and when a package is not directly available, I will just use pip to install. So one get around with this. I think if we have already have a fix and starting retraining the model, it is best to update this accordingly otherwise we might have to revisit this issue in the future which we will have to remind ourselves again of what this is about :D

R-Walmsley:
Hi @chanshing, This will only directly affect people who install via the installation instructions in the README.md - all the code is compatible with earlier versions of sklearn. But I've amended the requirements file to reflect that.

chanshing:
> > Hi Rosemary, I thought we were sticking to sklearn 0.23 for now since 0.24 is not yet available in conda? Anaconda is the most popular manager for data scientists and I'm afraid it might break for many users.
> 
> I think this is probably fine because I normally use Conda and when a package is not directly available, I will just use pip to install. So one get around with this. I think if we have already have a fix and starting retraining the model, it is best to update this accordingly otherwise we might have to revisit this issue in the future which we will have to remind ourselves again of what this is about :D

I'd prefer the path of least resistance, but if .24 works on .23 that's fine too. 
My idea was to stay on .23 but update the code and model. These will work just fine in .24 too (.24 just actually deprecates what it said was gonna deprecate for ages).

chanshing (approved):

R-Walmsley:
@chanshing Changes in the most recent commit should address your concerns - noone should look at the requirements.txt file (at least as far as sklearn goes) and not be able to install what it says because they're using conda (because I've changed it to '>= 0.23.0'). 

Separately, we discussed adding instructions for installation via conda, now raised as Issue #110 so I don't forget! 

I think this does get at a broader issue though. We've gone from having a very flexible specification of package versions, so open that the code broke, to a very rigid one, which we know works, but is very particular. We can gradually work to relax that, where we think it's important to do so. We can also think about it as 'here is one setup we know will work, any user is free to use other versions which are likely  to also work- but that's outside the scope of the checks/testing that we're doing'

R-Walmsley:
Hi @chanshing, @angerhang, 
I think this should now be ready to merge...

angerhang:
> Hi @chanshing, @angerhang,
> I think this should now be ready to merge...

Let's do this so that it would be one step closer to get the CI back to work.

chanshing:
It looks great! Please merge :dancing_men:

PR #112, opened 2021-01-27
Update reference epoch.csv to reflect changes introduced in 2fabe36

chanshing:
This is just to update the current referenceFiles/epoch.csv.gz to reflect changes introduced in 2fabe36 (handling of nan correlations). I should have done this right after that merge but forgot. The only difference in the csvs are in two rows (4902 & 13670) where some of `corrxy` and `corryz` are NaN in the current csv. After 2fabe36 they should now be 0.0.


chanshing:
@R-Walmsley @aidendoherty @angerhang
Let me shamelessly merge my own PR right away. There's no way to review this binary file anyways, and as you see the checks are passing now that Rosemary fixed the deprecation issues (https://github.com/activityMonitoring/biobankAccelerometerAnalysis/pull/108) :dancing_men:

PR #113, opened 2021-01-27
fix NA for avgArmAngelAbsDiff with epoch size 5sec

angerhang:
`avgArmAngelAbsDiff` should be ill-defined when one epoch has less than 10 sec as it is supposed to be the absolute difference between the angel z for two 5-sec window. Returning 0 to make for practical purpose instead of NA.

Fixed the issue mentioned in https://github.com/activityMonitoring/biobankAccelerometerAnalysis/issues/103.

### DIFF
--- a/java/Features.java
+++ b/java/Features.java
@@ -621,7 +621,9 @@ public class Features {
      * A novel, open access method to assess sleep duration using a wrist-worn accelerometer
      * van Hees et al. 2015
      *
-     * This won't work if the epoch length is less than 5-sec. So return 0 in that case.
+     * avgArmAngel is only defined for epoch_length >= 5 sec and avgArmAngelAbsDiff
+     * is only defined for epoch_length >= 10 sec. If epoch_length is not valid, we will return 0.
+     * For more details, please refer to the references above.
      */
     private static double[] calculateArmFeatures(double[] x,
                                                  double[] y,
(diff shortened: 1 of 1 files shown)


java/Features.java
```diff
@@ -651,10 +652,12 @@ private static String getUnileverFeaturesHeader(){
...
-            double[] absoluteAvgDiff = computeAbsoluteDiff(fiveSecAvg);
-
-            // get the avg of difference
-            double avgArmAngelAbsDiff = AccStats.mean(absoluteAvgDiff);
+            //    ill-defined when we have fewer 10 sec for the epoch length
+            double avgArmAngelAbsDiff = 0;
+            if (x.length/sampleRate >= 10) {
+                double[] absoluteAvgDiff = computeAbsoluteDiff(fiveSecAvg);
+                avgArmAngelAbsDiff = AccStats.mean(absoluteAvgDiff);
+            }
```
chanshing, line 660: Shall we also explain this at the top of the function? Currently it says

`This won't work if the epoch length is less than 5-sec. So return 0 in that case.`

If I understand correctly `avgArmAngel` requires >=5, but `avgArmAngelDiff` requires >=10 right?

chanshing (changes requested):
Thanks Hang. Hope my minor comment helps.

java/Features.java
angerhang: Yes that's correct. Okay let me perhaps explain it a bit more.

R-Walmsley:
Hi @angerhang , Looks fine to me - though I'm not at all familiar with this end of the code, or with Java...

chanshing:
Thanks Hang. Finally, do you mind squashing both commits? After that should be ready to go :dancing_men:

angerhang:
> Thanks Hang. Finally, do you mind squashing both commits? After that should be ready to go 👯‍♂️

Done ;D

chanshing:
Thanks Hang. Best practice on commit messages from what I read is to use imperative tense ( "fix X" instead of "fixed X", "update X" instead of "updated X") https://www.danclarke.com/git-tense
Just a minor thing in case you feel like you need to force-push someone.
Otherwise, feel free to merge :dancing_men:

chanshing (approved):

PR #117, opened 2021-02-06
Major refactoring: Pythonization of large parts of the processing code

chanshing:
Note: this is **still in progress**

As discussed, eventually we'd like to pythonize the whole codebase to make it easier for the research community (incl. ourselves) to contribute. This PR is to pythonize large part of the processing code e.g. frequency regularization (resampling), feature extraction, gravity calibration (TODO), nonwear detection (TODO), etc. The parsing of the device binary file is still done in Java (we support 3 devices, each with a different data format requiring different low-level parsing, so let's postpone this for now).

For now, the aim is to limit the role of Java to just parsing. Some untangling had to be done as the Java code is currently doing many things like resampling, feature extraction, quality checks, etc. all in between the parsing. A Java parser for the Axivity device has now been cleared (AxivityParser.java), and untangling the remaining two is coming soon (TODO: ActigraphParser.java, GENEActivParser.java).

#### Current flow
```
             device.py 
                |
    AccelerometerParser.java 
                |  \
                |   DeviceReader.setupEpochWriter()
                |  /
         XYZReader.java             # parsing
                |
      epochWriter.newValues()       # still parsing, resample & calibrate along the way
                |
   epochWriter.writeEpochSummary()  # still parsing, extract feats along the way
```

#### Proposed flow
```
             device.py 
                |
         XYZParser.java        # parsing
                |              # done parsing
      processing.Processing()  # resample, calibration, extract feats, etc.
```

### DIFF
--- a/accProcess.py
+++ b/accProcess.py
@@ -364,7 +364,7 @@ def main():
                     os.close(npyFileFd)
                     os.unlink(npyFile)
 
-            device.parse(
+            info = device.parse(
                 inputFile=args.inputFile,
                 outputFile=npyFile,
                 timeZone=args.timeZone,
--- a/accProcess.py
+++ b/accProcess.py
@@ -15,6 +15,7 @@ import warnings
 from accelerometer import device
 import numpy as np
 from accelerometer import processing
+from tempfile import mkstemp
 
 
 def main():
@@ -399,13 +400,25 @@ def main():
     #     "In total, processing took " + str(processingTime) + " seconds"
     # )
 
-    tempOutFile = device.parse(
+    if args.npyOutput:
+        npyFile = args.npyFile
+
+    else:
+        npyFileFd, npyFile = mkstemp()
+
+        @atexit.register
+        def closeTemp():
+            os.close(npyFileFd)
+            os.unlink(npyFile)
+
+    device.parse(
         args.inputFile,
+        npyFile,
         timeZone=args.timeZone,
         timeShift=args.timeShift
     )
 
-    processing.Processing(tempOutFile, args.sampleRate, args.epochPeriod, args.epochFile)
+    processing.Processing(npyFile, args.sampleRate, args.epochPeriod, args.epochFile)
 
 
 def str2bool(v):
--- a/accProcess.py
+++ b/accProcess.py
@@ -12,6 +12,9 @@ import accelerometer.summariseEpoch
 import pandas as pd
 import atexit
 import warnings
+from accelerometer import device
+import numpy as np
+from accelerometer import processing
 
 
 def main():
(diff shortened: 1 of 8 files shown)


chanshing:
As said above, PR still in progress but submitting it for partial feedback as I work on the rest -- the gist of it is already there. There is some untangling left to do for ActigraphParser.java and GENEActivParser.java and some plumbing with summariseEpoch.py and accClassification.py

accelerometer/processing.py
```diff
@@ -0,0 +1,202 @@
...
+MAXVAL, MINVAL = 6, -6  # Note: GENEActiv has a range of +/-6g whereas Axivity and Actigraph +/-8 or more
+
+
+class Processing():
+    def __init__(self, npyFile, sampleRate, epochPeriod, epochFile):
+        self.npyFile = npyFile
+        self.sampleRate = sampleRate
+        self.epochPeriod = epochPeriod
+        self.epochFile = epochFile
+
```
aidendoherty, line 22: Somewhere below, we would ideally also identify 'stuck' values i.e. where the device continually records a value for an unrealistically long time period.

See https://github.com/activityMonitoring/biobankAccelerometerAnalysis/blob/master/java/AccStats.java#L333

accelerometer/processing.py
```diff
@@ -0,0 +1,202 @@
...
+#TODO
+def detectNonwear():
+    pass
+
+
+def extractFeatures(data, sampleRate, epochPeriod):
+    epochs = data.groupby(pd.Grouper(freq=f'{epochPeriod}S'))
+
+    def extract(data):
```
aidendoherty, line 74: Do we really want a function within a function?

accelerometer/processing.py
```diff
@@ -0,0 +1,202 @@
...
+    feats['zxCorr'] = np.nan_to_num(np.corrcoef(z, x)[0,1])
+
+    m = np.linalg.norm(xyz, axis=1)
+    feats['enmoTrunc'] = np.mean(np.maximum(m - 1.0, 0.0))
+    feats['mad'] = stats.median_abs_deviation(m)
+
+    return feats
+
+
+def sanDiegoFeatures(xyz, sampleRate):
```
aidendoherty, line 106: It is great that you have faithfully reimplemented the San Diego and Unilever functions. We can test these for backwards compatibility. Soon after this pull request it would then be good to condense/simplify this feature set, probably in the work @angerhang has planned.

java/AxivityParser.java
```diff
@@ -0,0 +1,373 @@
...
+    private String outFile;
+    private String timeZone;
+    private int timeShift;
+
+    private ZonedDateTime sessionStart = null;
+    private boolean isSessionStartDST;
+    private ZoneId zoneId;
+    private ZoneRules rules;
+
+    private int errCounter = 0;
```
aidendoherty, line 31: Is there someway we can return this? Perhaps by returning a particular error value in the npy file so that it can then be captured in the subsequent python processing?

java/AxivityParser.java
```diff
@@ -0,0 +1,373 @@
...
+                        yRaw = 0;
+                        zRaw = 0;
+                    }
+                    x = xRaw / 256.0;
+                    y = yRaw / 256.0;
+                    z = zRaw / 256.0;
+                    t = zonedWithDSTCorrection(blockTime).toInstant().toEpochMilli();
+
+                    try {
+                        writer.writeData(t, (float) x, (float) y, (float) z);
```
aidendoherty, line 259: Would it be better practice to have writer passed in as an argument to this block or code (rather than being used as a global variable)?

java/NpyWriter.java
```diff
@@ -11,7 +11,7 @@
 
 public class NpyWriter {
 
-    private static final Boolean COMPRESS = true;
+    private static final boolean COMPRESS = false;
```
aidendoherty, line 14: Is this for processing speed reasons?
In case we want to keep the temporary .npy file (see first python comment) we would probably want it compressed. Will of course trust your good judgement on this.

aidendoherty (commented):
This looks very exciting @chanshing - what a huge effort!
I've added some comments throughout, but this is certainly going in the right direction and will much improve the underlying codebase.

accProcess.py
```diff
@@ -331,67 +334,75 @@ def deleteIntermediateFiles():
...
+    # ##########################
+    # # Closing
+    # ##########################
+    # processingEndTime = datetime.datetime.now()
+    # processingTime = (processingEndTime - processingStartTime).total_seconds()
+    # accelerometer.accUtils.toScreen(
+    #     "In total, processing took " + str(processingTime) + " seconds"
+    # )
+
+    tempOutFile = device.parse(
```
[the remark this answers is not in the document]
chanshing: > This looks very exciting @chanshing - what a huge effort!
> I've added some comments throughout, but this is certainly going in the right direction and will much improve the underlying codebase.

Thank you Aiden for the helpful feedback :)

chanshing:
> This looks very exciting @chanshing - what a huge effort!
> I've added some comments throughout, but this is certainly going in the right direction and will much improve the underlying codebase.

Thank you Aiden for the helpful feedback :)

accelerometer/processing.py
```diff
@@ -0,0 +1,202 @@
...
+        data = regularizeSampleRate(data, sampleRate)
+        data = filterNoise(data, sampleRate)
+        #TODO
+        # data = calibrateGravity(...)
+        #TODO
+        # data = detectNonwear(...)
+
+        epochFeats = extractFeatures(data, sampleRate, epochPeriod)
+
+        epochFeats.to_csv(epochFile)
```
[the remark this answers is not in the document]
chanshing: Thanks Aiden. I found out that pandas is clever enough to automatically infer compression from filename endings. Still I will change this to `to_csv(epochFile, compression='gzip')` -- explicit better than implicit :smiley:

accelerometer/processing.py
chanshing: Let me rethinkg about this

java/AxivityParser.java
chanshing: Yes, it's part of my TODO list. I was waiting to dig into Actigraph and GENEA to see what other metrics are returned by them and see if I can abstract them.

java/AxivityParser.java
chanshing: I thought about this and concluded that it'd be best to use an object-oriented pattern and make `writer` a private attribute of a parser object. In this view, a `writer` is an inherent part of a parser instance. This also saves us from having to pass `writer` to `parseCwaBuffer`, in fact, `writer` doesn't change at all during the loop and is only passed so that the function gets a hold of the pointer. In the process, I also made `errCounter`, `lastBlockTimeIndex`, and `lastBlockTime` private attributes of the parser object for similar reasons:

#### Current
```java
// AxivityReader.java

// Lines 36-66
// Although errCounter in practice acts as a simple int counter, it has to be defined as 
// a size 1 array so that we can pass it to readAccBuffer below and persist changes. 
// Same goes for lastBlockTime and lastBlockTimeIndex.
int[] errCounter = new int[] { 0 };
LocalDateTime[] lastBlockTime = { null };
int[] lastBlockTimeIndex = { 0 };

// ...
try {
    // ...
    while (rawAccReader.read(buf) != -1) {
        // Here the only thing that's worth passing is buf
        // The other variables are mostly bookkeeping variables
        readCwaBuffer(buf, USE_PRECISE_TIME, lastBlockTime, lastBlockTimeIndex, header, errCounter, epochWriter);
        buf.clear();
    }
} catch {
// ...
}

// Further down, inside readCwaBuffer()

// Because we had to pass errCounter, lastBlockTime, lastBlockTimeIndex as size 1 arrays,
// we get some strange looking lines for simple increments, assignments, comparisons:

// Line 240
if (lastBlockTime[0] != null) {
// ...
}

// Line 259
lastBlockTime[0] = blockTime;
lastBlockTimeIndex[0] = timestampOffset - sampleCount;

// Line 294
errCounter[0] += 1;
```
#### Proposed
```java
// AxivityParser.java -- this PR

// Using an object-oriented pattern, keep errCounter, lastBlockTime, lastBlockTimeIndex and writer as
// private variables of "this" object
private int errCounter = 0;
private LocalDateTime lastBlockTime = null;
private int lastBlockTimeIndex = 0;
private NpyWriter writer;

// In parse()

try {
    // ...
    while (accChannel.read(buf) != -1) {
        // Just passing buf
        parseCwaBuffer(buf);
        buf.clear();
    }
} catch {
// ...

// Further down, in parseCwaBuffer()
// Variables used intuitively
if (lastBlockTime[0] != null) {
// ...
}
// ...
lastBlockTime = blockTime;
lastBlockTimeIndex = timestampOffset - sampleCount;
// ...
errCounter += 1;
```

java/NpyWriter.java
chanshing: Thanks Aiden. Yes, fully for performance reasons. When I had implemented the compression I had included it as a default when `close()`ing. In hindsight, I think this is strange and I'll change it to make compression explicit.

java/AxivityParser.java
chanshing: Looking at ActigraphReader and GENEActiveReader, I'm considering reverting to the object-less, static-based pattern as otherwise it'll take too much time for little gain.

accProcess.py
```diff
@@ -331,67 +334,75 @@ def deleteIntermediateFiles():
...
+        args.inputFile,
```
angerhang, line 399: Regarding the use of parameters. Do you think it would be a good idea to replace Argpase with Hydra?

angerhang:
> Note: this is **still in progress**
> 
> As discussed, eventually we'd like to pythonize the whole codebase to make it easier for the research community (incl. ourselves) to contribute. This PR is to pythonize large part of the processing code e.g. frequency regularization (resampling), feature extraction, gravity calibration (TODO), nonwear detection (TODO), etc. The parsing of the device binary file is still done in Java (we support 3 devices, each with a different data format requiring different low-level parsing, so let's postpone this for now).
> 
> For now, the aim is to limit the role of Java to just parsing. Some untangling had to be done as the Java code is currently doing many things like resampling, feature extraction, quality checks, etc. all in between the parsing. A Java parser for the Axivity device has now been cleared (AxivityParser.java), and untangling the remaining two is coming soon (TODO: ActigraphParser.java, GENEActivParser.java).
> 
> #### Current flow
> ```
>              device.py 
>                 |
>     AccelerometerParser.java 
>                 |  \
>                 |   DeviceReader.setupEpochWriter()
>                 |  /
>          XYZReader.java             # parsing
>                 |
>       epochWriter.newValues()       # still parsing, resample & calibrate along the way
>                 |
>    epochWriter.writeEpochSummary()  # still parsing, extract feats along the way
> ```
> 
> #### Proposed flow
> ```
>              device.py 
>                 |
>          XYZParser.java        # parsing
>                 |              # done parsing
>       processing.Processing()  # resample, calibration, extract feats, etc.
> ```
Hey Shing thanks a lot for the nice PR. I've looked at the preliminary changes and found most of the appropriate. I just have some general questions understanding that this is a preliminary PR for feedback: 

1. How are the processing looking like in the proposed workflow to do resample, calibration and other things?
2. In terms of the change of parameters passing using Hydra. As we talked about, Hydra gives us more structured ways to manage input options. Not sure if the current refactoring work  will be a good time to introduce this.
3. In terms of the next steps, we might want to introduce some additional regression tests to ensure that our refactoring is not too invasive. I could probably contribute to that when this PR comes to a closer time for merging. One of the test is already breaking but that's because of a dependency issue. It would be interesting to see how we do the tests properly because obviously we won't have a perfect match in terms of data output even when the code is a `perfect` replicate of the java version :D

java/AxivityParser.java
aidendoherty: That sounds fine.

accProcess.py
aidendoherty: I've had a quick look at Hydra and it looks fabulous and very well suited to our needs. We should do this. However, perhaps it should be in our next pull request?
What do you think @chanshing ?

chanshing: Thank you both. Yes I think it'd be best to put it as a separate PR :smile:

chanshing:
> Hey Shing thanks a lot for the nice PR. I've looked at the preliminary changes and found most of the appropriate. I just have some general questions understanding that this is a preliminary PR for feedback:
> 
> 1. How are the processing looking like in the proposed workflow to do resample, calibration and other things?
> 2. In terms of the change of parameters passing using Hydra. As we talked about, Hydra gives us more structured ways to manage input options. Not sure if the current refactoring work  will be a good time to introduce this.
> 3. In terms of the next steps, we might want to introduce some additional regression tests to ensure that our refactoring is not too invasive. I could probably contribute to that when this PR comes to a closer time for merging. One of the test is already breaking but that's because of a dependency issue. It would be interesting to see how we do the tests properly because obviously we won't have a perfect match in terms of data output even when the code is a `perfect` replicate of the java version :D

Thanks Hang. Hydra is nice and I think it'd be better included separately as this PR is already huge enough.
Yes it'd be good to include a simple test and thanks for offering to help :smile:

chanshing:
@aidendoherty @angerhang @R-Walmsley 
Changes since last review:
1. Added ActigraphParser.java and GENEActivParser.java. I also made significant changes to NpyWriter.java so that one can customize what variables are to be written. For example, Axivity has temperature and lux but Actigraph doesn't (or at least it's not being parsed at the moment), so would be nice to be able to customize the writer object instead of the hardcoded time,x,y,z only. Note that temperature is used for calibration so I needed to implement the change to write this out. Also, it might be useful much later in case we start supporting say Axivity6 and 9 (with gyro, magneto, etc.)
2. Improved processing.py -- fixed some edge cases and hopefully is much cleaner now.

I still need to address some of the feedback but slowly getting there! 
Tons of commits but ofc I will rebase it all once it's OK to go :dancing_men:

accelerometer/device.py
```diff
@@ -9,6 +9,54 @@
...
+import shutil
+from tempfile import mkstemp
+import re
+
+
+
+def parse(inputFile, outputFile, **kwargs):
+
+    if inputFile.endswith('.gz'):  # Decompress to temporary file
+        ext = re.search(r'(\.\w+)\.gz', inputFile).group(1)
```
aidendoherty, line 23: Is this really needed?
The previous Java implementations were able to read the compressed .gz files ... an advantage of this is a reduced storage footprint (rather than having to create a temporary uncompressed file)

accelerometer/device.py
```diff
@@ -9,6 +9,54 @@
...
+
+        inputFile = _inputFile
+
+    # Start JVM
+    if not jpype.isJVMStarted():
+        jpype.addClassPath('../java/')
+        jpype.addClassPath('../java/JTransforms-3.1-with-dependencies.jar')
+        jpype.startJVM(convertStrings=False)
+
+    # Parsing -- parsed data is stored at tempOutFile
```
aidendoherty, line 38: I haven't read the rest of the code yet, but is there a reason this goes to tempOutFile rather than outputFile (the argument passed in to the relevant methods just below)?

accelerometer/processing.py
```diff
@@ -0,0 +1,356 @@
...
+        data = loadNpyToFrame(npyFile)
+
+        #TODO quality control
+
+        data = Processing.regularizeSampleRate(data, self.sampleRate)
+        data = Processing.filterNoise(data, self.sampleRate)
+
+        #TODO
+        # Used to be done in device.py but would make sense to do it here.
+        # data = calibrateGravity(...)
```
aidendoherty, line 39: Yes, I agree here would be a good place for this

accelerometer/processing.py
```diff
@@ -0,0 +1,356 @@
...
+
+        #TODO
+        # Currently done in summariseEpoch but would make more sense to do it
+        # here. Also the approach is slightly different to the paper.
+        # data = detectNonwear(...)
```
aidendoherty, line 44: We would also like this function to output the start/end times of each non-wear episode (to support later imputation)

java/GENEActivParser.java
```diff
@@ -0,0 +1,229 @@
...
+
+        int fileHeaderSize = 59;
+        int linesToAxesCalibration = 47;
+        int pageHeaderSize = 9;
+        int[] errCounter = new int[] { 0 };
+
+        NpyWriter writer = new NpyWriter(outFile, ITEM_NAMES_AND_TYPES);
+
+        try {
+            BufferedReader rawAccReader = new BufferedReader(new FileReader(accFile));
```
aidendoherty, line 45: It would be good to continue to have the current functionality of directly reading .gz files

aidendoherty (commented):
Many thanks for sharing @chanshing 
Still looks like it's moving in a great direction.

My only major comment is that it would be good to retain the functionality of reading direct from .gz files (as in current Java classes). For processing multiple files from larger studies, there is a big gain in being able to store and process these as compressed files

accelerometer/device.py
chanshing: Ah wrong comment. It is indeed outputFile. I have fixed it now.

accelerometer/device.py
chanshing: Thanks Aiden. It appears that we only got the functionality only for Axivity so far, so I went for the easiest approach. But yes I agree with you and will look into implementing .gz reads in Actigraph and GENEA.  

Captura de pantalla de 2021-02-13 08-06-22

chanshing: Thanks Aiden. It appears that we only got the [functionality only for Axivity so far](https://github.com/activityMonitoring/biobankAccelerometerAnalysis/blob/master/java/AccelerometerParser.java#L214-L230), so I went for the easiest approach. But yes I agree with you and will look into implementing .gz direct reads in Actigraph and GENEA.

chanshing:
Development will continue in https://github.com/activityMonitoring/pywear

PR #118, opened 2021-02-06
add feature set used in model Walmsley et al 2020 and remove old files

R-Walmsley:
This very simple PR is the long overdue adding of the feature set used in the most recent model. (Rotation invariant features). 

I also suggest deleting old files containing different feature sets- these use old feature names and erroneously contain some rotation-variant features. They can also be easily recreated according to the specific needs of specific projects. 

@smallsr - even before this is merged the file walmsley2020-features.txt might be useful for your project.


chanshing (approved):
Thanks Rosemary. Looks good!

PR #140, opened 2021-04-02
add support for nearest neighbor resampling

chanshing:
This adds support for nearest neighbor resampling (currently, only linear resampling is supported).
To use, set the new flag: `--resampleMethod [linear/nearest] (default: linear)`

Use cases: You may want to use the NN method if you want to be more elaborate with the preprocessing since linear resampling is a low pass filter (and we already provide that as an option separately)

How I tested it: I checked the output against pandas' `resample(...).nearest(1)` method and pandas' `pd.merge_asof(..., direction='nearest')` and they were consistent. Actually, there was an offset of 1 in the former, while it was exactly the same in the latter.


chanshing:
@smallsr sort of needs this ASAP so if you could review this soon he'd appreciate it.

angerhang (commented):
Looks good to me but if it is possible to include a test which uses the NN resampling. Since this is urgent, we don't need to verify the final output but ensure that a file a file can be processed at least?

chanshing:
> Looks good to me but if it is possible to include a test which uses the NN resampling. Since this is urgent, we don't need to verify the final output but ensure that a file a file can be processed at least?

Thanks Hang. Scott just independently tested it last night and it worked. May I *not* include a test right now? Reason being that it will change again very soon with the pythonization project. Also, you can follow these steps to test the PR yourself:  https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/checking-out-pull-requests-locally

Side comment: What should we do with the increasing check load? This will only keep going up with every new feature added.

angerhang:
> > Looks good to me but if it is possible to include a test which uses the NN resampling. Since this is urgent, we don't need to verify the final output but ensure that a file a file can be processed at least?
> 
> Thanks Hang. Scott just independently tested it last night and it worked. May I _not_ include a test right now? Reason being that it will change again very soon with the pythonization project. Also, you can follow these steps to test the PR yourself: https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/checking-out-pull-requests-locally
> 
> Side comment: What should we do with the increasing check load? This will only keep going up with every new feature added.

I see then it is probably ok for merge then. 

For the increasing check load, I think it is OK? We are not getting close to any thing that other large-scale open-source software. Pytorch for example will have to run through over 70+ CI tests for a pull request :D

aidendoherty:
I agree with @angerhang - it is fine for us to have many different/new checks and tests and for the set of tests to grow over time.

PR #142, opened 2021-04-04
fix warning pandas keep_tz

chanshing:
To shut this pandas deprecation warning:

```bash
FutureWarning: The 'keep_tz' keyword in DatetimeIndex.to_series is deprecated and will be removed in a future version.  You can stop passing 'keep_tz' to silence this warning.
  e_new.index = e_new.index.to_series(keep_tz=True).apply(date_strftime)
```

### DIFF
--- a/accelerometer/accUtils.py
+++ b/accelerometer/accUtils.py
@@ -147,8 +147,8 @@ def writeStudyAccProcessCmds(accDir, outDir, cmdsFile='processCmds.txt',
     nonWearDir = os.path.join(outDir, 'nonWear')
     stationaryDir = os.path.join(outDir, 'stationary')
     logsDir = os.path.join(outDir, 'clusterLogs')
-    rawDir = os.path.join(outDir, 'raw') 
-    npyDir = os.path.join(outDir, 'npy') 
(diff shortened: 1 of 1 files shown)


R-Walmsley:
So, if I understand correctly, the timezone is now kept by default? 

I guess the only risk is people using a very old version of `pandas` lose the timezone? But this is unlikely, and probably not a big issue?

chanshing:
> So, if I understand correctly, the timezone is now kept by default?
> 
> I guess the only risk is people using a very old version of `pandas` lose the timezone? But this is unlikely, and probably not a big issue?

Yeah apparently in old pandas it defaulted to `False` but in 1.0.0 they decided to default it to `True` https://pandas.pydata.org/docs/reference/api/pandas.DatetimeIndex.to_series.html
Not only that, they will remove it entirely (while defaulting to `True`) in the future.
Since we're requiring people to use pandas 1.1.5 I think this shouldn't be a problem https://github.com/activityMonitoring/biobankAccelerometerAnalysis/blob/5431429ec1450e7fd01f0a2133028ae3f14f8e16/requirements.txt#L16

R-Walmsley:
Agreed!

PR #156, opened 2021-05-06
Added support for Windows

cosmicnet:
Hi,

The following has been tested on Windows 10 with conda 4.10.1


angerhang:
Thanks for the PR @cosmicnet . After a quick glance, changes in this PR is grossly similar to the changes that I made in the DREAMS trial for windows support with better documentation. Likely to work. 

Guess @chanshing @R-Walmsley would be the best people to review this since you have direct access to windows machines?

chanshing:
@cosmicnet Hi Lyle, do we need to updated the README in this PR? Package is in PyPI now so we now install using `pip install accelerometer`.

cosmicnet:
Yeah, a lot has changed since I did this. I can update it to work with the new code. I think all that is needed in the readme now would be a note to say that the windows support is experimental. Unless you have other ideas?

chanshing:
Fixed in 9a26001c356d51933f53fd43e0e47a5cb36d045f
Thanks @cosmicnet !

cosmicnet:
Thanks for fixing this!

Issue #158, opened 2021-05-10
Small mistake in root mean square calculation in calibration code


chanshing:
There's a small, perhaps negligible, mistake in calculating the root mean square:
https://github.com/activityMonitoring/biobankAccelerometerAnalysis/blob/0d977b904ca076b05d7ffae4ccabc2eecd32babb/accelerometer/device.py#L270

Should be `np.sqrt(np.mean(np.square(np.linalg.norm(curr-target, axis=1))))`

chanshing:
BTW it'd be worth using instead median absolute error (simply `np.median(np.linalg.norm(curr-target, axis=1))`) which is more robust to outliers (in case we accidentatly grab a nonstationary window for calibration).

PR #160, opened 2021-05-17
fix small bug in calibration rmse calculation

chanshing:
Closes https://github.com/activityMonitoring/biobankAccelerometerAnalysis/issues/158

This is to fix a small mistake in RMSE calculation in the calibration code as mentioned in https://github.com/activityMonitoring/biobankAccelerometerAnalysis/issues/158

### DIFF
--- a/accelerometer/device.py
+++ b/accelerometer/device.py
@@ -266,8 +266,9 @@ def getCalibrationCoefs(staticBoutsFile, summary):
     bestTemp = np.copy(tempCoef)
     # record initial uncalibrated error
     curr = intercept + (np.copy(axesVals) * slope) + (np.copy(tempVals) * tempCoef)
-    target = curr / np.sqrt(np.sum(np.square(curr), axis=1))[:, None]
-    initError = np.sqrt(np.mean(np.square(curr-target)))  # root mean square error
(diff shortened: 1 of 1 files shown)


R-Walmsley (commented):
Hi @chanshing 

Might be useful to add a comment on what this does? 

The only difference I can currently see is that the call to `np.mean` now seems to not make sense as it's now on a 1d array. But I think I may be misunderstanding the dimensions of the input (and thought asking would be easier than chasing it up the tree :) )

chanshing:
> Hi @chanshing
> 
> Might be useful to add a comment on what this does?
> 
> The only difference I can currently see is that the call to `np.mean` now seems to not make sense as it's now on a 1d array. But I think I may be misunderstanding the dimensions of the input (and thought asking would be easier than chasing it up the tree, though happy to chase if the answer isn't straightforward! :) )

Hi, yes so `curr` and `target` are 2D arrays of shape (N,3) representing the triaxial signal.
Currently we do `initError = np.sqrt(np.mean(np.square(curr-target)))  # root mean square error` so the mean here is being taken across both temporal and axial dimensions, so it is off by a factor 3 from true mean. What we want is `np.sqrt(np.mean(np.square(np.linalg.norm(curr-target, axis=1))))`.

PR #161, opened 2021-05-24
print git hash used in accProcess

R-Walmsley:
A first attempt at printing some of the details of processing as part of the summary. Feedback very welcome! (And please don't merge yet, see below for why). 

TODO: 
- Test in different settings (e.g. installed/non-installed). 
- This only does it for accProcess.py, which means: 1. It doesn't actually tell you which version of accelerometer has been used for different processing (although according to the rules for finding python modules, it should be this one. 2. It doesn't help with any of the other functions in the package (that's a "TODO list" item). 
- The regression test will evidently need updating!
- Probably worth testing for multiple environments as I guess it might be a bit sensitive to that sort of thing.

### DIFF
--- a/accProcess.py
+++ b/accProcess.py
@@ -12,6 +12,7 @@ import accelerometer.summariseEpoch
 import pandas as pd
 import atexit
 import warnings
+import subprocess
 
 
 def main():
@@ -336,6 +337,8 @@ def main():
     # Start processing file
     ##########################
     summary = {}
+    
+    
     # Now process the .CWA file
     if args.processInputFile:
         summary['file-name'] = args.inputFile
@@ -374,6 +377,9 @@ def main():
     # Generate time series file
     accelerometer.accUtils.writeTimeSeries(epochData, labels, args.tsFile)
 
+    # Record git hash used for this processing 
+    summary['git-hash'] = str(subprocess.check_output(["git", "describe"]).strip())
+   
     # Print short summary
     accelerometer.accUtils.toScreen("=== Short summary ===")
     summaryVals = ['file-name', 'file-startTime', 'file-endTime',


aidendoherty:
Looks fine to me at face value. For full reproducibility/logging, would we also want to output the input parameters too?

R-Walmsley:
Hi @aidendoherty 

Yes in theory. That said, they are more usually stored because typically there is the text file with commands, so probably not the priority. 

Actually this is pretty far from full reproducibility, as all this is checking is the git hash of the accProcess.py file used. I guess in theory that could end up calling a different version of `accelerometer` (although I think only if the user had done something funky with the file structure). 

So maybe how exactly this should work is something for a quick discussion in one of our biobankAccelerometerAnalysis meetings?

angerhang:
> Hi @aidendoherty
> 
> Yes in theory. That said, they are more usually stored because typically there is the text file with commands (although defaults won't be obvious from this), so maybe not the priority.
> 
> Actually this is pretty far from full reproducibility, as all this is checking is the git hash of the accProcess.py file used. I guess in theory that could end up calling a different version of `accelerometer` (although I think only if the user had done something funky with the file structure).
> 
> So maybe how exactly this should work is something for a quick discussion in one of our biobankAccelerometerAnalysis meetings?

Being able to know which version of BBAA is certainly great! Thanks for the initiatives :D 

In fact once we move to using Hydra as part of the pythonization move, all the input parameters will automatically saved so we don't need to spend time doing something that will be better done at a later date.

In terms of this PR, what is the expected behaviour if the command fails? Sometimes the end users will have Git installed. 

Yeah we could potentially discuss this really quick in the meeting but printing out the Github hash is the least that we can do. Better than nothing right even though not bullet-proof ;D

chanshing:
Closing as we now version our packages in Pip

PR #165, opened 2021-06-24
new feature to flexibly read in raw .csv files

aidendoherty:
new feature to flexibly read in raw .csv files:
 Time,x,y,z
OR
 Time,x,y,z,temperature

This will not be used very much, but is helpful in some edge case scenarios (e.g. when testing new accelerometer devices).

### DIFF
--- a/accProcess.py
+++ b/accProcess.py
@@ -13,7 +13,6 @@ import pandas as pd
 import atexit
 import warnings
 
-
 def main():
     """
     Application entry point responsible for parsing command line requests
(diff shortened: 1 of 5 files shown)


accProcess.py
```diff
@@ -1,5 +1,6 @@
 """Command line tool to extract meaningful health info from accelerometer data."""
 
+import dxpy
```
chanshing: Is this part of the PR?

chanshing (commented):
Thanks Aiden.
Is the dxpy package part of the PR?

accProcess.py
aidendoherty: Thanks @chanshing, I can confirm this is not part of the PR. My last two minor commits were used to rectify this. Let me see if there is a way that I can squash these into the first commit...

PR #166, opened 2021-06-25
Plot update

aidendoherty:
This pull request makes two updates/improvements to the plotting function:

 1) Removes annoying requirement to specify output plot filename (it can still be specified of course on calling the --plotFile flag

 2) Adds --showFileName flag to toggle if user would like to display the file name in the plot title or not (helpful for some edge cases)

### DIFF
--- a/accPlot.py
+++ b/accPlot.py
@@ -71,7 +71,7 @@ def main():
     args = parser.parse_args()
 
     # determine output file name
-    if args.plotFile == None:
+    if args.plotFile is None:
         inputFileFolder, inputFileName = os.path.split(args.timeSeriesFile)
         inputFileName = inputFileName.split('.')[0]  # remove any extension
         args.plotFile = os.path.join(inputFileFolder, inputFileName + "-plot.png")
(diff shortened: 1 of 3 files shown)


aidendoherty:
I think there may be an error in the testing framework with respect to python versions 3.6 and 3.7?

accPlot.py
```diff
@@ -61,14 +62,21 @@ def main():
...
+    if len(sys.argv) < 2:
+        msg = "\nInvalid input, please enter at least 1 parameter, e.g."
+        msg += "\npython accPlot.py timeSeries.csv.gz \n"
         accUtils.toScreen(msg)
         parser.print_help()
         sys.exit(-1)
     args = parser.parse_args()
 
+    # determine output file name
+    if args.plotFile == None:
```
chanshing, line 74: Can we change to `if args.plotFile is None:` which is more pythonic?

chanshing:
I like this change. Thanks Aiden!

accPlot.py
aidendoherty: Many thanks @chanshing - I've now updated this based on your helpful suggestion

aidendoherty:
I've also added a final feature to add capability to show just the first N days of a plot --showFirstNDays

PR #171, opened 2021-07-19
fix bug which before read the packetLength field rather than the device…

aidendoherty:
…Id field. Now working again


aidendoherty:
I'm going to integrate this for a collaborator. I have tested it on a few different Axivity devices to verify the output is correct. We can look to setup a formal test for this within the pywear framework.

chanshing:
Thanks @aidendoherty 
I'll integrate this change into pywear as well.

PR #185, opened 2021-11-17
Improve docs

R-Walmsley:
Some changes to docs as discussed with @aidendoherty this morning. 

- Add a note that different versions of the model and tool are not necessarily the same. (I'm particularly thinking of scenario where people use current versions of models e.g. in a particular patient population and compare them with UKB returns results. I think  models aren't comparable in that way, particularly due to the changes to the underlying data - notably around sleep).
- Remove some instructions around downloading models which I think are now redundant. 
- Update the citation of the BJSM paper from the preprint to the real version.  

Should these changes be approved, I will squash them before committing :)

### DIFF
--- a/docs/source/methods.rst
+++ b/docs/source/methods.rst
@@ -4,17 +4,18 @@ Methods
 
 Interpreted levels of physical activity can vary, as many approaches can be 
 taken to extract summary physical activity information from raw accelerometer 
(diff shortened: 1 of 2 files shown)


R-Walmsley:
@aidendoherty @chanshing can we merge this? 

Docs aren't currently building anyway - not currently sure why. See: https://readthedocs.org/projects/biobankaccanalysis/builds/?page=1

However, if @chanshing is also updating docs will be useful to have this merged in?

chanshing:
Thanks Rosemary 👯‍♂️ !

PR #200, opened 2022-06-15
fix tz for gt3x

angerhang:
Having spoken with the Actigrpah team, they have confirmed that all the timestamps are stored in UNIX UTC, the true UNIX time. The current expected behaviour is that:

* For npy file, we will just read the timestamps are they are.
* If a user provides the tz flag, we will change the epoch tz to the user input one. Otherwise we will default to `Europe/London`.

This PR should address: issue https://github.com/OxWearables/biobankAccelerometerAnalysis/issues/198

### DIFF
--- a/accelerometer/accProcess.py
+++ b/accelerometer/accProcess.py
@@ -43,6 +43,12 @@ def main():  # noqa: C901
                             -15 will shift the device internal time by -15
                             minutes. Not to be confused with timezone offsets.
                             (default : %(default)s""")
+    parser.add_argument('--offsetHour',
+                        metavar='e.g. 00:00:00', default='00:00:00',
+                        type=str, help="""offsetHour is used by GT3X V1 only.
+                        When processing a file not using UTC, make sure 
+                        to specify the hour offset from UTC.
+                            (default : %(default)s""")
     parser.add_argument('--startTime',
                         metavar='e.g. 1991-01-01T23:59', default=None,
                         type=str2date, help="""removes data before this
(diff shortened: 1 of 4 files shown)


aidendoherty:
Looks good to me but we should also ask @chanshing to review too.

I wonder if this also need to be done for GeneActiv and Axivity devices?
We can discuss this in person when we next meet in the office.

angerhang:
> Looks good to me but we should also ask @chanshing to review too.
> 
> I wonder if this also need to be done for GeneActiv and Axivity devices? We can discuss this in person when we next meet in the office.

Thanks for the comment Aiden. All three devices deal with time zone differently. But I plan to submit a PR later this week to include a minimal test set at least for GT3X and Axivity devices. 

As reported in https://github.com/OxWearables/biobankAccelerometerAnalysis/issues/199, there seems to be an issue with Axivity time zone parsing as well but I have left that to Shing to fix as I am less familiar with its encoding.

angerhang:
After some investigation, GT3X's UNIX time is in fact local time. The user will need to provide both the time shift information in hours and timezone to get the correct output. If nothing is provided, we will default everything to `Europe/London`.

angerhang:
@chanshing can you review this quickly and publish a new release when you have the time? I kinda need to get this done soon for a bunch of analyses that I have to run :D

It is probably time to set up automatic PR release for pip distribution.

chanshing:
Hang, I think we don't need offset + timezone. Timezone should be enough. If device is only storing local time, then I think it is very similar to Axivity (and will have same problem regarding DST)

chanshing:
This might be useful: https://docs.microsoft.com/en-us/dotnet/api/system.datetime.ticks?view=net-6.0
It says that .NET ticks stores ticks since Jan 0001 midnight in either UTC or a local time zone. What we need now is the inverse transform of this: f(ticks, timezone) -> time (unix, UTC, or any localized time).

angerhang:
I have spent the whole morning trying to do this. Even writing the test cases took a long time and I still haven't finished one test case yet probably due to my lack of familiarity with Java and its timezone classes. GT3X has two versions add to the complexity. Can we just merge this PR? It is better than the current version already because the current version is just sometimes wrong because of the timezone.

To get this correct with the right set of test cases, it will take 2-3 days' time. I have to prioritise other tasks at the moment.
We really should move away from merging PR without test cases. The more we build on things, the more we question if things are really going to work.

angerhang:
Anyhow, I will probably work with my own fork for now. Maybe revisit this at a later date.

Issue #204, opened 2022-07-22
accPlot bug when provided timeseries has gaps


chanshing:
![857a1f41-4022-4a0d-b716-57eb734cdcda](https://user-images.githubusercontent.com/5551642/180433565-be7fac63-87e8-4aec-9677-d88e86b1e8f8.jpg)

Should be easy to fix by using `.reindex(new_index, method='nearest', tolerance='1m')` or similar.

Quick manual fix gave:
![86b96dc9-fd14-456b-a9bd-b9e2dfa24c63](https://user-images.githubusercontent.com/5551642/180433620-b9170795-0250-44fd-aa43-61c6c76eb1c9.jpg)

Thanks @gmertes for catching it.

PR #205, opened 2022-07-22
Add fillGaps argument for when input has gaps (#204)

gmertes:
Simple workaround for this issue. Will fill gaps in the input time series with `np.NaN` so they correctly show up as grey areas in the plot. Downside with using `pd.reindex` is that it's not doing a "real" resampling. It still expects the input have an exact uniform sampling rate (and the `fillGaps` arg value needs to match this exactly). 

We could also infer the sample interval from the input, but I think that is more prone to errors down the line. Since this is probably a niche issue, probably best to put it behind a manual switch?

Usage:
`accPlot timeSeries.csv --fillGaps 30s`

### DIFF
--- a/accelerometer/accPlot.py
+++ b/accelerometer/accPlot.py
@@ -56,12 +56,6 @@ def main():  # noqa: C901
     parser.add_argument('--showFirstNDays',
                         metavar='days', default=None,
                         type=int, help="Show just first n days")
-    parser.add_argument('--fillGaps',
-                        metavar='freq',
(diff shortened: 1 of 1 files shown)


chanshing:
Thank you so much, Gert! 

I made a few modifications based on your solution: 

1. Infer the freq from data or use 30s as default. This should cover most of our use cases.
2.  A bit more forgiving resampling so that it will work for both 1, 2, 3, --, 5, 6, 7, but also 1, 2, 3, --, 5.1, 6.1, 7.1, etc.

PR #206, opened 2022-07-28
refactor accPlot.plotTimeSeries

gmertes:
Refactor `plotTimeSeries` to a general function that creates the pyplot Figure from a DataFrame input. User can then decide what to do with it (show, save, manipulate it further, etc). Useful for when your data is already in memory and you don't want/need to convert to csv first. Also allows you to easily display the figure in a notebook or interactive shell.

Functionality and interface of original accPlot cmdline program remains unchanged. 

Usage: 
```
from accelerometer.accPlot import plotTimeSeries
df = pd.DataFrame(...)
fig = plotTimeSeries(df)
fig.show()
```

Example in notebook (here with a preloaded DataFrame for simplicity, but the main use case is that the DF can be constructed on the fly from any source): 
image

Original usage test:
```
python accPlot.py sample-timeSeries.csv
<Plot file written to: sample-timeSeries-plot.png>
```
Output: [sample-timeSeries-plot.png](https://user-images.githubusercontent.com/13658335/181587661-2af81604-8f83-4d7f-9168-24c651ad31ff.png)

Note:
- The change from `ax.set_title` to `ax.set_ylabel` was necessary because the day labels would fall out of the figure. Calling `savefig `with `bbox_inches='tight'` fixes this, but only on the saved file and not on the Figure object itself. `fig.tight_layout()` is supposed to fix this, but it doesn't (maybe a bug in matplotlib?). Using the axis label instead of the title solved it. There's no functional difference other than the horizontal alignment now being on the right. Advantage of this is that the label is now auto padded on the left side of the axis, resulting in less wasted white-space between label and axis.

### DIFF
--- a/accelerometer/accPlot.py
+++ b/accelerometer/accPlot.py
@@ -12,7 +12,6 @@ from datetime import datetime, timedelta, time
 import argparse
 from accelerometer import utils
 import matplotlib
-matplotlib.use('Agg')
 
 # http://pandas-docs.github.io/pandas-docs-travis/whatsnew/v0.21.1.html#restore-matplotlib-datetime-converter-registration
 register_matplotlib_converters()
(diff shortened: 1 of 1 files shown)


chanshing:
Amazing Gert, thank you so much!

PR #209, opened 2022-08-31
speed up missing data imputation

gmertes:
Speed up missing data imputation by a factor ~4. Improvement comes from:

- `DataFrameGroupBy.transform` has the unique behaviour of first passing the group subframe column-by-column (as a Series) to the provided function, and then again in its entirety (as a DataFrame). If it gets a valid DataFrame return value, it will pass subsequent chunks as DataFrames only. This can be beneficial if your transform function is faster on the whole subframe, but in our case:
- It means that with the original implementation, after the first chunk, `x` is a DataFrame, `x.mean()` is also a DataFrame, so `x.fillna(x.mean())` is doing a conditional join of 2 DataFrames. This is expensive. The mean is also calculated for each input even when there are no missing values.
- It is much faster to do `fillna` with a scalar value, and we could do that column-by-column. But also:
- Numpy vectorisation is faster than Pandas' indexing, and converting a Pandas Series to Numpy is cheap. Since our input data type is float, there's no issue converting the Series to a Numpy array and back.

So: we rewrite the transform function so that it only works column-by-column and use Numpy vectorisation to replace NaN values, which results in a significant speed up.

``` python
from accelerometer.summarisation import imputeMissing

# old
%time imputed = imputeMissing(original)
CPU times: total: 30.1 s
Wall time: 30.1 s

# new
%time imputed = imputeMissing(original)
CPU times: total: 7.45 s
Wall time: 7.49 s
```

To show that it does the same thing, here is an example imputed time series plotted with accPlot (original, old imputed and new imputed). All three imputation cases (day of week, weekday/weekend, all other days) are done correctly.

image

image

image

### DIFF
--- a/accelerometer/summarisation.py
+++ b/accelerometer/summarisation.py
@@ -259,17 +259,31 @@ def imputeMissing(data, extrapolate=True):
             tolerance=pd.Timedelta('1m'),
             limit=1)
 
+    def fillna(subframe):
+        # Transform will first pass the subframe column-by-column as a Series.
+        # After passing all columns, it will pass the entire subframe again as a DataFrame.
+        # Processing the entire subframe is optional (return value can be omitted). See 'Notes' in transform doc.
+        if isinstance(subframe, pd.Series):
+            x = subframe.to_numpy()
+            nan = np.isnan(x)
+            nanlen = len(x[nan])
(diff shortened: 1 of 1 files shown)


chanshing:
Amazing work @gmertes Thank you so much!
I pulled and verified that it replicates the results at a much faster speed 🚀

PR #218, opened 2022-12-01
Extract light from ax3

angerhang:
* Include lux extraction from ax3 to epoch file
* Add dummy light variable to device readers without light

Test plan:
```
python accelerometer/accProcess.py data/hang.cwa --deleteIntermediateFiles False
```

Output:
Screenshot 2022-12-01 at 19 44 53

To do in a different PR, add summary statistics from the epoch files


aidendoherty:
Would it be better to introduce this enhancement in Actipy rather than BBA @angerhang @chanshing ?

chanshing:
Thanks, Hang. I'll put it on my todo list to implement it in actipy also.

chanshing:
All looks good. I'll think about how to rename `ax3_light` and to distinguish it from `light`. Thanks a lot Hang!

angerhang:
This feature is put on hold until the actipy backend migration finishes sometime next month.

Will revisit this end of January.

